PR #4, opened 2021-07-07
Feature/add-the-crosscompilation-support

dchvs:


rrcarlosrodriguez (changes requested):

README.md
```diff
@@ -1,6 +1,23 @@
...
+## Build the GstTIOVX with cross compilation
+
+```
+# Add the path to the cross compiler
+CROSS_COMPILE=<...>
+# Add the path to the targetfs/
+SYSROOT_PATH=<...>
+
+# Update the Meson's cross compile file with the cross compiler & sysroot
+sed -i "s|cross_compiler=''|cross_compiler='$CROSS_COMPILE'|g" crosscompile.txt
```
rrcarlosrodriguez, line 13: I don't like the user doing this every time. We should find a way to do this automatically.

rrcarlosrodriguez (changes requested):

README.md
```diff
@@ -1,6 +1,23 @@
 # GstTIOVX
 Repository to host GStreamer plugins for TI's EdgeAI class of devices
 
+## Build the GstTIOVX with cross compilation
```
michaelgruner, line 4: Cross compiling the project

README.md
```diff
@@ -1,6 +1,23 @@
...
+CROSS_COMPILE=<...>
+# Add the path to the targetfs/
+SYSROOT_PATH=<...>
+
+# Update the Meson's cross compile file with the cross compiler & sysroot
+sed -i "s|cross_compiler=''|cross_compiler='$CROSS_COMPILE'|g" crosscompile.txt
+sed -i "s|sysroot_path=''|sysroot_path='$SYSROOT_PATH'|g" crosscompile.txt
+
+# Build in cross compilation mode
+meson build --prefix=/usr --cross-file crosscompile.txt
```
michaelgruner, line 17: The idea of being able to specify which cross-compile file to use, is that eventually we have several files for different architectures. I recommend naming this file after the architecture (the prefix in the toolchain we use, i.e: aarch64-none-linux) and moving them to a `cross` directory.

crosscompile.txt
```diff
@@ -0,0 +1,34 @@
+[constants]
+cross_compiler=''
```
michaelgruner, line 2: Given that this will be specific to a platform, we can hardcode this to TI's toolchain.

crosscompile.txt
```diff
@@ -0,0 +1,34 @@
...
+sysroot_path=''
+
+[binaries]
+c = cross_compiler + 'gcc'
+cpp = cross_compiler + 'g++'
+ld = cross_compiler + 'ld'
+ar = cross_compiler + 'ar'
+strip = cross_compiler + 'strip'
+pkgconfig = '/usr/bin/pkg-config'
```
michaelgruner, line 11: Don't force any specific binary on the user. Use plain `pkg-config`  or `/usr/bin/env pkg-config`

crosscompile.txt
```diff
@@ -0,0 +1,34 @@
...
+cpu = 'aarch64'
+endian = 'little'
+
+[built-in]
+sys_root = sysroot_path
+pkg_config_sysroot_dir = sysroot_path
+pkg_config_path = sysroot_path + '/usr/lib/pkgconfig/'
+
+prefix = '/usr'
+libdir = 'lib/'
```
michaelgruner, line 30: This is wrong I think. It should be `prefix + '/lib/`

meson.build
```diff
@@ -19,19 +19,21 @@ api_version = '1.0'
 
 # Required versions
 gst_req = '>= 1.0.0'
+glib_req = '>= 2.0'
 
 # Find external dependencies
 gst_dep       = dependency('gstreamer-1.0',      version : gst_req)
 gst_base_dep  = dependency('gstreamer-base-1.0', version : gst_req)
 gst_check_dep = dependency('gstreamer-check-1.0',version : gst_req)
+glib_dep      = dependency('glib-2.0',           version : glib_req)
```
michaelgruner, line 28: I would expect that gstreamer already carries this as a dependency?

michaelgruner (changes requested):

pkgconfig/tiovx.pc
```diff
@@ -0,0 +1,20 @@
+prefix=/usr
+exec_prefix=${prefix}
+libdir=${prefix}/lib/
+includedir=${prefix}/include/processor_sdk/tiovx/include/
+includedir1=${prefix}/include/processor_sdk/tiovx/kernels_j7/include/
```
michaelgruner, line 5: I like this, but I don't like the names. Let's use significative names, for example: `includedir_j7kernels`, `includedir_kernels`, `includedirs_utils`, etc...

michaelgruner (changes requested):

README.md
dchvs: Thanks. I added an environment to set up all the cross compilation required variables.

README.md
```diff
@@ -1,12 +1,39 @@
...
-ninja -C build/ install
+# Load the PSDKR path
+PSDKR_PATH=<...>
+
+# Create an environment with the:
+* Source path
+* Build path
+* PSDKR path
+
+chmod +x crossbuild/environment
```
emmadrigal, line 15: Can't you push this file already with the +x?

Also, since you are sourcing it, do you need to be an executable?

README.md
```diff
@@ -1,12 +1,39 @@
...
+mkdir build && cd build
+source ../crossbuild/environment  $PWD/.. $PWD  $PSDKR_PATH
+
+# Build in cross compilation mode
+meson .. --prefix=$PWD/deploy/gst-tiovx --cross-file crossbuild/aarch64.ini --cross-file ../crossbuild/crosscompile.txt
+ninja -j8
```
emmadrigal, line 21: You don't need to force ninja to use multiple cores it will already attempt to use the max amount: https://ninja-build.org/manual.html#_running_ninja.

Leave it as `ninja`

README.md
```diff
@@ -1,12 +1,39 @@
...
+dpkg -x deploy/gst-tiovx.deb $PSDKR_PATH/targetfs/usr/
+
+```
+
+## Compiling the project native
+
+```
+mkdir build && cd build
+meson .. --prefix=/usr -Dpkg_config_path=../pkgconfig
+ninja  -j8
```
emmadrigal, line 35: See my comment above about multiple cores

README.md
```diff
@@ -1,12 +1,39 @@
...
+ninja -j8
+ninja install
+
+# Deploy the binaries
+ninja debfile
+dpkg -x deploy/gst-tiovx.deb $PSDKR_PATH/targetfs/usr/
+
+```
+
+## Compiling the project native
```
emmadrigal, line 30: Typo `natively`

README.md
```diff
@@ -1,12 +1,39 @@
...
+## Cross compiling the project
 
 ```
-meson build --prefix=/usr
-ninja -C build/ -j8
-ninja -C build/ install
+# Load the PSDKR path
+PSDKR_PATH=<...>
+
+# Create an environment with the:
```
emmadrigal, line 10: I don't think its clear what you mean by this...

Is this:
* sourcing the directories
* creating said directories
* something else?

crossbuild/DEBIAN/control
```diff
@@ -0,0 +1,5 @@
+Package: gst-ti-ovx
+Version: 0.1.0
+Maintainer: Daniel Chaves <daniel.chaves@ridgerun.com>
```
emmadrigal, line 3: I'm not sure if you should be the maintainer, in other projects we have used the client's info.

Please ask this in our internal chat team.

gst-libs/gst/tiovx/gsttiovx.c
```diff
@@ -64,3 +64,11 @@
 #endif
 
 #include "gsttiovx.h"
+
+#include "app_color_convert_module.h"
```
emmadrigal, line 68: Wrong include order: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

pkgconfig/vision_apps_modules.pc
```diff
@@ -0,0 +1,27 @@
...
+includedirs_ivision=${prefix}/include/processor_sdk/ivision/
+includedirs_imaging=${prefix}/include/processor_sdk/imaging/
+
+toolsdir=${exec_prefix}/bin
+pluginsdir=${prefix}/lib/
+datarootdir=${prefix}/share
+datadir=${datarootdir}
+
+Name: TIOVX
+Description: TI OVX API
```
emmadrigal, line 22: Wrong name and description

emmadrigal (changes requested):

README.md
dchvs: Thanks, it's not necessary to make it executable; for some reason it just seemed to me. Change applied.

crossbuild/environment
```diff
@@ -0,0 +1,18 @@
...
+PSDKR_PATH=$3
+
+CROSS_COMPILER=$PSDKR_PATH/gcc-arm-9.2-2019.12-x86_64-aarch64-none-linux-gnu/bin/
+PSDKR_SYSROOT=$PSDKR_PATH/targetfs/
+PKGCONFIG=$SOURCE_ROOT/pkgconfig/
+
+PATH=$PATH:$CROSS_COMPILER
+
+mkdir $BUILD_ROOT/crossbuild/
+cat > $BUILD_ROOT/crossbuild/aarch64.ini <<EOL
```
m-herrera, line 13: I think it's safer to surround variables with {} as well as ""
for example `cat > "${BUILD_ROOT}"/crossbuild/aarch64.ini <<EOL`

Same applies for all the variables in  this file

m-herrera (changes requested):

emmadrigal (approved):

PR #8, opened 2021-07-15
Feature/add gsttiovxsimo class

jafet-chaves:


gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+
+  ret = TRUE;
+
+  return ret;
+}
+
+static gboolean
+gst_tiovx_simo_query (GstPad * pad, GstObject * parent, GstQuery * query)
+{
+  GstTIOVXSimo *self;
```
emmadrigal, line 485: Init to `NULL`

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  gboolean ret = FALSE;
+
+  self = GST_TIOVX_SIMO (parent);
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  switch (GST_QUERY_TYPE (query)) {
+    case GST_QUERY_CAPS:
+    {
+      GstCaps *filter;
```
emmadrigal, line 497: Init to `NULL`

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  GstCaps *outcaps = NULL;
+  GstTIOVXSimoClass *klass = NULL;
+  GstTIOVXSimoPrivate *priv = NULL;
+  vx_status status = VX_FAILURE;
+  gboolean ret = FALSE;
+  GstVideoInfo in_info;
+  GstVideoInfo out_info;
+  vx_graph_parameter_queue_params_t *params_list;
+  guint parameter_index = 0;
+  gpointer pool_size;
```
emmadrigal, line 564: Init to NULL

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+    GST_ERROR_OBJECT (self, "Incomplete info from subclass: node missing");
+    goto free_graph;
+  }
+  if (0 == priv->num_pads) {
+    GST_ERROR_OBJECT (self,
+        "Incomplete info from subclass: number of graph parameters is 0");
+    goto free_graph;
+  }
+
+  params_list = malloc (priv->num_pads * sizeof (*params_list));
```
emmadrigal, line 653: Use `g_malloc0` and release with `g_free`

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+
+exit:
+  return ret;
+}
+
+static gboolean
+gst_tiovx_simo_sink_event (GstPad * pad, GstObject * parent, GstEvent * event)
+{
+  GstTIOVXSimoPrivate *priv = NULL;
+  GstTIOVXSimo *self;
```
emmadrigal, line 741: init to `NULL`

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  gboolean ret = FALSE;
+
+  self = GST_TIOVX_SIMO (parent);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  switch (GST_EVENT_TYPE (event)) {
+    case GST_EVENT_CAPS:
+    {
+      GstCaps *caps;
```
emmadrigal, line 750: Init to null

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+
+      /* TODO: call fixate_caps virtual method first */
```
emmadrigal, line 752: Why is there a `TODO` here?

emmadrigal (changes requested):

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  g_return_val_if_fail (priv->node, VX_FAILURE);
+
+  graph = priv->graph;
+  node = *priv->node;
+
+  parameter = vxGetParameterByIndex (node, parameter_index);
+  status = vxAddParameterToGraph (graph, parameter);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Add parameter to graph failed");
+    return status;
```
dchvs, line 216: Please, use the `goto label `statements.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  appCommonDeInit ();
+
+  priv->init_completed = FALSE;
+
+exit:
+  return ret;
+}
+
+static void
+gst_tiovx_simo_finalize (GObject * object)
```
dchvs, line 305: `gobject`, might be a more significant variable.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  switch (transition) {
+      /* "Start" transition */
+    case GST_STATE_CHANGE_NULL_TO_READY:
+      gst_tiovx_simo_modules_init (self);
+      break;
+      /* "Stop" transition */
+    case GST_STATE_CHANGE_READY_TO_NULL:
+      gst_tiovx_simo_modules_deinit (self);
+      break;
+    default:
```
dchvs, line 344: Should we return `GST_STATE_CHANGE_FAILURE` here?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  status =
+      add_graph_parameter_by_node_index (self, parameter_index, params_list,
+      priv->input, priv->in_pool_size);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Setting input parameter failed");
+    goto free_parameters_list;
+  }
+
+  /*Starts on 1 since input parameter was already set */
+  for (parameter_index = 1; parameter_index < priv->num_pads; parameter_index++) {
```
dchvs, line 668: Please consider using the regular counter `i`, as it increases the code readability.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+    default:
+      ret = gst_pad_query_default (priv->sinkpad, parent, query);
+      break;
+  }
+
+  return ret;
+}
+
+static gboolean
+gst_tiovx_simo_set_caps (GstTIOVXSimo * self, GstPad * pad, GstCaps * incaps)
```
dchvs, line 553: This function is too big, please factorize it.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  vx_graph_parameter_queue_params_t *params_list;
+  guint parameter_index = 0;
+  gpointer pool_size;
+
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_DEBUG_OBJECT (pad, "have new caps %p %" GST_PTR_FORMAT, incaps, incaps);
+
+  if (!gst_video_info_from_caps (&in_info, incaps))
```
dchvs, line 571: You might want to use `g_assertions` here.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+  } else {
+    ret = klass->configure_module (self);
+    if (!ret) {
+      GST_ERROR_OBJECT (self, "Subclass configure node failed");
+      goto free_graph;
+    }
+  }
+
+  ret = TRUE;
+  goto exit;
```
dchvs, line 716: Follow the single-input/single-output pattern, that is, a natural sequence of the code until reaching out the return. So, the first "goto label" you step with from top to bottom, should be the last one to be declare and therefore sequeantilly freed.

gst-libs/gst/tiovx/gsttiovxsimo.h
```diff
@@ -0,0 +1,138 @@
...
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef _GST_TIOVX_SIMO_H_
+#define _GST_TIOVX_SIMO_H_
+
+#include <gst/gst.h>
+#include <gst/base/base-prelude.h>
```
dchvs, line 66: Alphabetical order.

dchvs (changes requested):
@jafet-chaves , changes requested

meson.build
```diff
@@ -26,6 +26,7 @@ gst_base_dep  = dependency('gstreamer-base-1.0', version : gst_req)
 gst_video_dep = dependency('gstreamer-video-1.0', version : gst_req)
 gst_check_dep = dependency('gstreamer-check-1.0',version : gst_req)
 gst_allocator = dependency('gstreamer-allocators-1.0', version : gst_req)
+gst_video_dep = dependency('gstreamer-video-1.0',version : gst_req)
```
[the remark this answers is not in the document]
jafet-chaves: Yes, this is needed but I see its duplicated. I'll remove it.

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: This is due to it was agreed this PR won't contain the implementation for this part.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,766 @@
...
+static gboolean gst_tiovx_simo_sink_event (GstPad * pad, GstObject * parent,
+    GstEvent * event);
+
+guint
+gst_tiovx_simo_get_num_pads (GstTIOVXSimo * self)
+{
+  GstTIOVXSimoPrivate *priv = NULL;
+  guint result;
+
+  g_return_val_if_fail (GST_IS_TIOVX_SIMO (self), FALSE);
```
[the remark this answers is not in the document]
jafet-chaves: GST_IS_TIOVX_SIMO already does it

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: No, I don't believe so this is required. That would mean the code fails for any other state transition.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,823 @@
...
+      GST_DEBUG_FUNCPTR (gst_tiovx_simo_sink_event));
+  gst_pad_set_query_function (priv->sinkpad,
+      GST_DEBUG_FUNCPTR (gst_tiovx_simo_query));
+  gst_element_add_pad (GST_ELEMENT (self), priv->sinkpad);
+}
+
+static vx_status
+add_graph_parameter_by_node_index (GstTIOVXSimo * self,
+    vx_uint32 parameter_index, vx_graph_parameter_queue_params_t params_list[],
+    vx_reference * handler, guint pool_size)
```
michaelgruner, line 228: is `handler` a list of size `pool_size`? If so, it is not a very good name. Use a name that hints the reader that `handler` is, in fact,  a list of `pool_size`. Maybe `ref_list` and `ref_list_size` or something like that.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,823 @@
...
+  GstTIOVXSimoPrivate *priv = NULL;
+  vx_status status = VX_FAILURE;
+  vx_parameter parameter = NULL;
+  vx_graph graph = NULL;
+  vx_node node = NULL;
+
+  g_return_val_if_fail (self, VX_FAILURE);
+  g_return_val_if_fail (handler, VX_FAILURE);
+  g_return_val_if_fail (parameter_index >= 0, VX_FAILURE);
+  g_return_val_if_fail (pool_size >= MIN_POOL_SIZE, VX_FAILURE);
```
michaelgruner, line 239: The problem I have with this is that this method seems generic (for any parameter), but here you are assuming that the parameter will be buffers.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,823 @@
...
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+  g_return_val_if_fail (priv, VX_FAILURE);
+  g_return_val_if_fail (priv->graph, VX_FAILURE);
+  g_return_val_if_fail (priv->node, VX_FAILURE);
+
+  graph = priv->graph;
+  node = *priv->node;
```
michaelgruner, line 247: This is another point that is confusing to me. If this is a list of nodes, why do we always assume that the parameter will be written to the first element in the list.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,823 @@
...
+
+  parameter = vxGetParameterByIndex (node, parameter_index);
+  status = vxAddParameterToGraph (graph, parameter);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Add parameter to graph failed");
```
michaelgruner, line 252: We need to free the parameter if this failed.

michaelgruner: Also, always put the status code of the failure.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,823 @@
...
+
+static gboolean
+gst_tiovx_simo_modules_init (GstTIOVXSimo * self)
+{
+  GstTIOVXSimoPrivate *priv = NULL;
+  guint index = 0;
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_DEBUG ("gst_ti_ovx_simo_modules_init");
```
michaelgruner, line 279: _OBJECT

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  params_list[parameter_index].refs_list_size = pool_size;
+  params_list[parameter_index].refs_list = (vx_reference *) handler;
+
+  status = VX_SUCCESS;
+
+  return status;
+}
+
+static gboolean
+gst_tiovx_simo_modules_null_to_ready_init (GstTIOVXSimo * self)
```
michaelgruner, line 272: start and stop are the common names for these methods. `gst_tiovx_simo_start` and `gst_tiovx_simo_stop`

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+  priv->in_pool_size = DEFAULT_POOL_SIZE;
+  priv->out_pool_sizes = g_hash_table_new (NULL, NULL);
+
+  for (index = 0; index < priv->num_pads; index++) {
+    g_hash_table_insert (priv->out_pool_sizes, GUINT_TO_POINTER (index),
+        GUINT_TO_POINTER (DEFAULT_POOL_SIZE));
+  }
+
+  priv->srcpads = g_hash_table_new (NULL, NULL);
```
michaelgruner, line 289: Wait, above you are asking for the number of pads, but here you are creating the hash table for the pads? this doesn't seem right. were are you storing the pads? pads are going to be created mostly in the null state, so by now we are already late.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  GstTIOVXSimoPrivate *priv = NULL;
+  vx_status status = VX_FAILURE;
+  gboolean ret = FALSE;
+  vx_graph_parameter_queue_params_t *params_list = NULL;
+  guint i = 0;
+  gpointer pool_size = NULL;
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  if (0 != appCommonInit ()) {
```
michaelgruner, line 309: Dont use the numeric value. Use the enum name.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  if (!ret) {
+    GST_ERROR_OBJECT (self, "Subclass init module failed");
+    goto free_context;
+  }
+
+  priv->graph = vxCreateGraph (priv->context);
+  status = vxGetStatus ((vx_reference) priv->graph);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Graph creation failed");
+    goto free_context;
```
michaelgruner, line 343: We need to call deinit_module here.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  }
+
+  if (!klass->create_graph) {
+    GST_ERROR_OBJECT (self, "Subclass did not implement create_graph method.");
+    goto free_graph;
```
michaelgruner, line 348: This single-input/single-output pattern is not okay. At this point you still don't have a graph. Please revise the full thing.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+gst_tiovx_simo_modules_init (GstTIOVXSimo * self, GstCaps * sink_caps,
+    GList * src_caps_list)
+{
+  GstTIOVXSimoClass *klass = NULL;
+  GstTIOVXSimoPrivate *priv = NULL;
+  vx_status status = VX_FAILURE;
+  gboolean ret = FALSE;
+  vx_graph_parameter_queue_params_t *params_list = NULL;
+  guint i = 0;
+  gpointer pool_size = NULL;
```
michaelgruner, line 305: Why is this a gpointer?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Setting input parameter failed");
+    goto free_parameters_list;
+  }
+
+  /*Starts on 1 since input parameter was already set */
+  for (i = 1; i < priv->num_pads; i++) {
+    if (g_hash_table_contains (priv->out_pool_sizes, GUINT_TO_POINTER (i))) {
+      pool_size =
+          g_hash_table_lookup (priv->out_pool_sizes, GUINT_TO_POINTER (i));
```
michaelgruner, line 404: Please use GPOINTER_TO_GUINT and make pool_size an integer.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+    } else {
+      GST_ERROR_OBJECT (self, "Fail to obtain output pool size");
+      goto free_parameters_list;
+    }
+
+    status =
+        add_graph_parameter_by_node_index (self, i, params_list,
+        priv->output[i], *(guint *) (pool_size));
```
michaelgruner, line 412: using pool_size as this is incorrect.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+  status = vxSetGraphScheduleConfig (priv->graph,
+      VX_GRAPH_SCHEDULE_MODE_QUEUE_MANUAL, priv->num_pads, params_list);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Graph schedule configuration failed");
+    goto free_parameters_list;
+  }
+
+  /* Parameters list has to be released even if the code doesn't fail */
+  g_free (params_list);
```
michaelgruner, line 427: is there a max number of potential parameters? if so, I would very much recommend using static memoryof dynamic allocation.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+        "Subclass did not implement configure node method. Skipping node configuration");
+  } else {
+    ret = klass->configure_module (self);
+    if (!ret) {
+      GST_ERROR_OBJECT (self, "Subclass configure node failed");
+      goto free_graph;
+    }
+  }
+
+  priv->module_init = TRUE;
```
michaelgruner, line 446: Do we really need another flag? can we use the graph or something like that?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+  ret = TRUE;
+  goto exit;
+
+free_parameters_list:
+  free (params_list);
+
+free_graph:
+  vxReleaseGraph (&priv->graph);
```
michaelgruner, line 455: Always null out freed variables.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+}
+
+static gboolean
+gst_tiovx_simo_modules_deinit (GstTIOVXSimo * self)
+{
+  GstTIOVXSimoPrivate *priv = NULL;
+  GstTIOVXSimoClass *klass = NULL;
+  gboolean ret = FALSE;
+
+  GST_DEBUG ("gst_ti_ovx_simo_modules_deinit");
```
michaelgruner, line 477: _OBJECT

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  gboolean ret;
+
+  self = GST_TIOVX_SIMO (gobject);
+
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+  gstelement_class = GST_ELEMENT_CLASS (klass);
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  g_hash_table_unref (priv->out_pool_sizes);
```
michaelgruner, line 532: This is not the place for this. You created the hash table on the start method, free it on the stop method.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  g_hash_table_remove_all (priv->srcpads);
+  g_hash_table_unref (priv->srcpads);
+
+  /* Handle case in which transition from READY to NULL did not occurred */
```
michaelgruner, line 536: This is not necessary if allocation/deallocation is done symmetrically.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+    case GST_STATE_CHANGE_NULL_TO_READY:
+      ret = gst_tiovx_simo_modules_null_to_ready_init (self);
+      if (!ret) {
+        GST_DEBUG_OBJECT (self,
+            "Failed to init module in NULL to READY transition");
+        return GST_STATE_CHANGE_FAILURE;
+      }
+      break;
+      /* "Stop" transition */
+    case GST_STATE_CHANGE_READY_TO_NULL:
```
michaelgruner, line 571: This is not the correct way to do this. You need to chain to the parent method. Please refer to the an example in the mainstream repos.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+    }
+  }
+
+  priv->module_init = TRUE;
+
+  ret = TRUE;
+  goto exit;
+
+free_parameters_list:
+  free (params_list);
```
michaelgruner, line 452: g_free

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  g_hash_table_insert (priv->srcpads, GUINT_TO_POINTER (index), srcpad);
+  priv->num_pads++;
+
+  g_free (name);
+
+  GST_OBJECT_UNLOCK (self);
+
+  gst_element_add_pad (GST_ELEMENT_CAST (self), srcpad);
+
+  return NULL;
```
michaelgruner, line 628: Why are you returning NULL?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  self = GST_TIOVX_SIMO (element);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_DEBUG_OBJECT (self, "requesting pad");
+
+  GST_OBJECT_LOCK (self);
+
+  if (name_templ && (sscanf (name_templ, "src_%u", &index) == 1)) {
+    GST_LOG_OBJECT (element, "name: %s (index %d)", name_templ, index);
+    if (g_hash_table_contains (priv->srcpads, GUINT_TO_POINTER (index))) {
```
michaelgruner, line 605: Are we testing this? how does this pass? you create the hash table after we start transition states, this should definitely fail.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  g_return_val_if_fail (src_caps_list, FALSE);
+
+  /* Loop through the list of src pads caps and by default do the following:
+   * Step 1. If there is a filter used it to intersect with the sink caps
+   * Step 2. Use the intersection obtained before to define the src caps
+   * and remove the width and height (this base class should not fix the
+   * width/height for each src pad) while keeping the rest as is.
+   */
+
+  sink_caps = gst_pad_get_pad_template_caps (sink_pad);
```
michaelgruner, line 675: You don't need this, the template is already considered.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+  priv->num_pads--;
+
+  GST_OBJECT_UNLOCK (self);
+
+  return;
+}
+
+static gboolean
+gst_tiovx_simo_default_get_caps (GstTIOVXSimo * self, GstPad * sink_pad,
```
michaelgruner, line 654: You don t need to pass in the pad

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+      "filter and sink caps intersected %" GST_PTR_FORMAT,
+      filter_sink_intersect);
+
+  while (NULL != src_caps_list) {
+    GstCaps *src_caps = NULL;
+    GstStructure *caps_struct = NULL;
+    GList *next = src_caps_sublist->next;
+
+    filtered_sink_src_intersect =
+        gst_caps_intersect_full (src_caps, filter_sink_intersect,
```
michaelgruner, line 690: What is going on here? src_caps are always null....

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+      gst_caps_intersect_full (filter, sink_caps, GST_CAPS_INTERSECT_FIRST);
+
+  GST_DEBUG_OBJECT (sink_pad,
+      "filter and sink caps intersected %" GST_PTR_FORMAT,
+      filter_sink_intersect);
+
+  while (NULL != src_caps_list) {
+    GstCaps *src_caps = NULL;
+    GstStructure *caps_struct = NULL;
+    GList *next = src_caps_sublist->next;
```
michaelgruner, line 687: always use g_list_next instead

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+    filtered_sink_src_intersect =
+        gst_caps_intersect_full (src_caps, filter_sink_intersect,
+        GST_CAPS_INTERSECT_FIRST);
+    GST_DEBUG_OBJECT (sink_pad,
+        "src and filtered sink caps intersected %" GST_PTR_FORMAT,
+        filtered_sink_src_intersect);
+
+    g_list_find (src_caps_sublist, (GstCaps *) filtered_sink_src_intersect);
```
michaelgruner, line 696: No this is incorrect. Just intersect all caps in the incoming list. no need for anything fancier than that.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+
+    /* Loop through the structures in GstCaps and remove height/width fields */
+    for (i = gst_caps_get_size (src_caps) - 1; i >= 0; --i) {
+      caps_struct = gst_caps_get_structure (src_caps, i);
+      gst_structure_remove_fields (caps_struct, "width", "height", NULL);
```
michaelgruner, line 701: Why are you doing this? the default is to intersect. This is a task for the multiscaler.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  priv->num_pads--;
+
+  GST_OBJECT_UNLOCK (self);
+
+  return;
+}
+
+static gboolean
+gst_tiovx_simo_default_get_caps (GstTIOVXSimo * self, GstPad * sink_pad,
+    GstCaps * filter, GList * src_caps_list)
```
michaelgruner, line 655: Where are you using the filter?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  gboolean ret = FALSE;
+
+  self = GST_TIOVX_SIMO (parent);
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  switch (GST_QUERY_TYPE (query)) {
+    case GST_QUERY_CAPS:
+    {
+      GstCaps *filter = NULL;
```
michaelgruner, line 733: Please move to a separate method.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+      if (pad != priv->sinkpad) {
+        GST_ERROR_OBJECT (self, "Pad from query is not the element sink pad");
+        return FALSE;
+      }
+
+      src_pads_list = g_hash_table_get_values (priv->srcpads);
+
+      src_pads_sublist = src_pads_list;
+      while (NULL != src_pads_sublist) {
+        GList *next = src_pads_sublist->next;
```
michaelgruner, line 750: Use g_list_next always.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+        GstPad *src_pad = NULL;
+
+        g_list_find (src_pads_sublist, (GstPad *) src_pad);
```
michaelgruner, line 753: This is not doing anything....

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+        /* Insert at the end of the src caps list */
+        src_caps_list =
+            g_list_insert (src_caps_list,
+            gst_pad_get_pad_template_caps (src_pad), -1);
```
michaelgruner, line 757: No, you need to send a caps query to the pad.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+    gst_caps_unref (filtered_sink_src_intersect);
+    gst_caps_unref (src_caps);
+  }
+
+  gst_caps_unref (sink_caps);
+  gst_caps_unref (filter_sink_intersect);
+
+  ret = TRUE;
+
+  return ret;
```
michaelgruner, line 715: Where are you returning the output caps? The method should return the output caps

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+      ret = klass->get_caps (self, pad, filter, src_caps_list);
+      if (!ret) {
+        GST_ERROR_OBJECT (self, "Get caps method failed");
+        return ret;
+      }
+
+      /* The query response should be the supported caps in the src pads except
+       * the width/height fields which is the only members that should differ
+       * from each src pad. */
+      gst_query_set_caps_result (query, g_list_nth_data (src_caps_list, 0));
```
michaelgruner, line 772: No, you should be adding here the caps returned by get caps...

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,838 @@
...
+  return ret;
+}
+
+static gboolean
+gst_tiovx_simo_set_caps (GstTIOVXSimo * self, GstPad * pad, GstCaps * sink_caps,
+    GList * src_caps_list)
+{
+  gboolean ret = FALSE;
+
+  GST_DEBUG_OBJECT (pad, "have new caps %p %" GST_PTR_FORMAT, sink_caps,
```
michaelgruner, line 794: What is going on here?

michaelgruner (changes requested):

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: No, handler is not supposed to be a list, instead is just an item from priv->output_refs. In any I agree handler is not a good name, perhaps "reference"?

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: What do you mean? appCommonInit is not part of the OpenVX API.

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: Yes, but it really depends of the OpenVX node instantiated in the subclass which the base class is not aware of beforehand.

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: Do you suggest to check pool_size >= MIN_POOL_SIZE outside this function, then?

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: I don't think it ever would be a case where priv->node is set to a list of vx_nodes in the first?

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: I think module_init works fine, there are other issues that have to be considered in using the graph. For example, the graph can be NULL and init_module have been called already (thus requiring to call deinit_module)

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+  g_return_val_if_fail (priv->node, VX_FAILURE);
+
+  graph = priv->graph;
+  node = priv->node;
+
+  parameter = vxGetParameterByIndex (node, parameter_index);
+  status = vxAddParameterToGraph (graph, parameter);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self,
+        "Add parameter to graph failed, vx_status %" G_GINT32_FORMAT,
```
michaelgruner, line 267: Normally, the explicit `G_GINT32_FORMAT` is only needed for 64 bits, because it is not portable to smaller 32bit platforms. In this case, `%d` is enough

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+        (int) status);
```
michaelgruner, line 268: I suggest you accustom to avoid casts unless explicitly needed. Specially in C, casts have supper powers and you can change anything into anything, which may be very dangerous. Try to remove all casts except for the ones that have a special reason to be there.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self,
+        "Release parameter failed, vx_status %" G_GINT32_FORMAT, (int) status);
+    return status;
+  }
+
+  params_list[parameter_index].graph_parameter_index = parameter_index;
+  params_list[parameter_index].refs_list_size = pool_size;
+  params_list[parameter_index].refs_list =
+      (vx_reference *) image_reference_list;
```
michaelgruner, line 283: Remove unnecessary cast

gst-libs/gst/tiovx/gsttiovxsimo.c
michaelgruner: I mean, does appCommonInit return 0 or something like VX_SUCCESS? if it has a named constant, use that.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+free_parameters_list:
+  g_free (params_list);
+
+free_graph:
+  vxReleaseGraph (&priv->graph);
+  priv->graph = NULL;
+
+deinit_module:
+  if (!klass->deinit_module) {
+    GST_ERROR_OBJECT (self, "Subclass did not implement deinit_module method");
```
michaelgruner, line 485: This walks straight into a segfault

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+  GstTIOVXSimoPrivate *priv = NULL;
+
+  self = GST_TIOVX_SIMO (gobject);
+
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+  gstelement_class = GST_ELEMENT_CLASS (klass);
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  g_hash_table_remove_all (priv->srcpads);
```
michaelgruner, line 571: Do we need to unref the pads?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+
+  srcpad = gst_pad_new_from_template (templ, name);
+  g_hash_table_insert (priv->srcpads, GUINT_TO_POINTER (index), srcpad);
+  priv->num_pads++;
+
+  g_free (name);
+
+  GST_OBJECT_UNLOCK (self);
+
+  gst_element_add_pad (GST_ELEMENT_CAST (self), srcpad);
```
michaelgruner, line 667: I believe we need to activate the pad

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+
+  self = GST_TIOVX_SIMO (element);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_OBJECT_LOCK (self);
+
+  gst_element_remove_pad (GST_ELEMENT_CAST (self), pad);
+
+  g_object_unref (pad);
+  g_hash_table_remove (priv->srcpads, GUINT_TO_POINTER (index));
```
michaelgruner, line 686: How does this compile? Where is index declared?

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  src_pads_list = g_hash_table_get_values (priv->srcpads);
+
+  src_pads_sublist = src_pads_list;
+  while (NULL != src_pads_sublist) {
+    GList *next = g_list_next (src_pads_sublist);
+
+    src_pad = (GstPad *) g_list_nth_data (src_pads_sublist, i);
```
michaelgruner, line 717: Use `GST_PAD()` safe cast. Remember the O(n) vs O(n^2) trick.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+
+  /* Loop through the list of src pads caps and by default fully
+   * intersect the list of source caps with the filter */
+  src_caps_sublist = src_caps_list;
+  ret = filter;
+  while (NULL != src_caps_list) {
+    GstCaps *src_caps = NULL;
+    GList *next = g_list_next (src_caps_sublist);
+
+    src_caps = (GstCaps *) src_caps_list->data;
```
michaelgruner, line 755: This is a very nasty bug. This should be sublist. Otherwise you'll get the same caps every time.

gst-libs/gst/tiovx/gsttiovxsimo.c
michaelgruner: ping

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,893 @@
...
+      src_caps_list = gst_tiovx_simo_get_src_caps_list (self, filter);
+      if (NULL == src_caps_list) {
+        GST_ERROR_OBJECT (self, "Get src caps list method failed");
+        return ret;
+      }
+
+      /* Should return the caps the element supports on the sink pad */
+      sink_caps = klass->get_caps (self, filter, src_caps_list);
+      if (NULL == sink_caps) {
+        GST_ERROR_OBJECT (self, "Get caps method failed");
```
michaelgruner, line 803: leak. Apply single-input/single-output

michaelgruner (changes requested):

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,876 @@
...
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifdef HAVE_CONFIG_H
+#  include "config.h"
+#endif
+
+#include "gsttiovxsimo.h"
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
jafet-chaves: To import sscanf in `gst_tiovx_simo_request_new_pad`

gst-libs/gst/tiovx/gsttiovxsimo.c
jafet-chaves: No if I create the hash table with `g_hash_table_new_full` and provide g_object_unref as the `value_destroy_func`: https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-new-full.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+  }
+  ret =
+      klass->init_module (self, priv->context, sink_caps, src_caps_list,
+      priv->in_pool_size, priv->out_pool_sizes);
+  if (!ret) {
+    GST_ERROR_OBJECT (self, "Subclass init module failed");
+    goto free_context;
+  }
+
+  priv->module_init = TRUE;
```
michaelgruner, line 354: This is misplaced here. If something goes wrong on the remainder of the function, this will remain TRUE, which will break the de-initialization.  I suggest dropping the extra flag and just use `priv->graph` as a initialization control flag.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+  gchar *name = NULL;
+
+  self = GST_TIOVX_SIMO (element);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_DEBUG_OBJECT (self, "requesting pad");
+
+  GST_OBJECT_LOCK (self);
+
+  name = g_strdup_printf ("src_%u", priv->num_pads);
```
michaelgruner, line 641: 1. This will break the case where the user passes in a named pad (such as `src_0` instead of `src_%d`).
2. Using num_pads to name the pad is wrong. Imagine the following scenario:

```
                                   : num_pads = 0
request_pad (src_0) : num_pads = 1
request_pad (src_1) : num_pads = 2
 remove_pad (src_0): num_pads = 1
request_pad (src_1) <----- ERROR
```
remove_pad (src_0)

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+{
+  GstTIOVXSimo *self = NULL;
+  GstTIOVXSimoPrivate *priv = NULL;
+
+  self = GST_TIOVX_SIMO (element);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_OBJECT_LOCK (self);
+
+  gst_element_remove_pad (GST_ELEMENT_CAST (self), pad);
```
michaelgruner, line 676: I'm unsure if you can call this with the LOCK held. Consider this if you get a deadlock during tests.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+
+  self = GST_TIOVX_SIMO (element);
+  priv = gst_tiovx_simo_get_instance_private (self);
+
+  GST_DEBUG_OBJECT (self, "requesting pad");
+
+  GST_OBJECT_LOCK (self);
+
+  name = g_strdup_printf ("src_%u", priv->num_pads);
+  srcpad = gst_pad_new_from_template (templ, name);
```
michaelgruner, line 642: Check if this returns a valid pad. In case of error, srcpad will be null.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+  while (NULL != src_pads_sublist) {
+    GstPad *src_pad = NULL;
+    GList *next = g_list_next (src_pads_sublist);
+
+    src_pad = GST_PAD (src_pads_sublist->data);
+    g_return_val_if_fail (src_pad, NULL);
+
+    /* Ask peer for what should the source caps (sink caps in the other end) be */
+    peer_caps = gst_pad_peer_query_caps (src_pad, filter);
+    g_return_val_if_fail (peer_caps, NULL);
```
michaelgruner, line 715: Be careful of mixing runtime errors with programming errors. `g_return_val_if_fail` is for programming errors. I believe this is okay, but be mindful about the difference.

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,890 @@
...
+  src_caps_sublist = src_caps_list;
+  ret = filter;
+  while (NULL != src_caps_list) {
+    GstCaps *src_caps = NULL;
+    GList *next = g_list_next (src_caps_sublist);
+
+    src_caps = (GstCaps *) src_caps_sublist->data;
+    g_return_val_if_fail (src_caps, NULL);
+
+    ret = gst_caps_intersect_full (ret, src_caps, GST_CAPS_INTERSECT_FIRST);
```
michaelgruner, line 751: This produces a leak in `ret`. By overwriting the `ret` the last reference is lost. always unref the original `ret` after the intersection.

michaelgruner (changes requested):

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,924 @@
...
+    while (NULL != key_sublist) {
+      GList *next = g_list_next (key_sublist);
+
+      index++;
+
+      key_sublist = next;
+    }
+
+    priv->next_pad_index = index + 1;
+  }
```
michaelgruner, line 661: This logic is not okay. This is way more complicated than needed. Consider:
```
If name is src_%d:
 - Walk through pads and get highest index
 - Increase index by 1
 - Use the increased index
if name is src_N (for an integer N):
 - Walk through pads and search if N exists
 - if not, create the pad
```

michaelgruner (changes requested):

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -0,0 +1,924 @@
...
+  gboolean ret = FALSE;
+
+  GST_DEBUG_OBJECT (self, "gst_ti_ovx_simo_modules_deinit");
+
+  g_return_val_if_fail (self, FALSE);
+
+  priv = gst_tiovx_simo_get_instance_private (self);
+  klass = GST_TIOVX_SIMO_GET_CLASS (self);
+
+  if (VX_SUCCESS != vxGetStatus ((vx_reference) priv->graph)) {
```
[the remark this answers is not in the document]
jafet-chaves: vxGetStatus fails if graph is NULL

emmadrigal (changes requested):

emmadrigal (approved):

PR #9, opened 2021-07-16
Feature/add multiscaler element

dchvs:


ext/tiovx/gsttiovxmultiscaler.h
```diff
@@ -0,0 +1,87 @@
...
+                     TIOVX_MULTI_SCALER, GstTIOVXSimo)
+
+G_END_DECLS
+
+#define MIN_POOL_SIZE 2
+#define MAX_POOL_SIZE 16
+
+#ifndef HOST_EMULATION
+#define HOST_EMULATION 0
+#endif
```
michaelgruner, line 84: Do we need this to be public? if not, move to the source.

tests/check/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,187 @@
...
+};
+
+GST_START_TEST (test_playing_to_null_multiple_times)
+{
+  test_states_change (test_pipelines[TEST_PLAYING_TO_NULL_MULTIPLE_TIMES]);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_init_module)
```
michaelgruner, line 101: I don't understand this test very much. What are we testing?

michaelgruner (changes requested):

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,458 @@
...
+  multiscaler->input.height = (vx_int32) GST_VIDEO_INFO_HEIGHT (in_info);
+  multiscaler->method = VX_INTERPOLATION_BILINEAR;
+  multiscaler->input.color_format = (vx_int32)
+      gst_tiovx_utils_map_gst_video_format_to_vx_format (in_info->
+      finfo->format);
+  multiscaler->input.bufq_depth = (vx_int32) in_pool_size;
+
+  /* Initialize the output parameters */
+  for (i = 0; i < multiscaler->num_outputs; i++) {
+    multiscaler->output[i].width = (vx_int32) GST_VIDEO_INFO_WIDTH (out_info);
```
jafet-chaves, line 324: This has to be fixed, you are using the same width and height for all outputs. This also requires to modify the gst_tiovx_multi_scaler_init_module declaration, probably using a list of output caps instead of out_info.

dchvs: Thank you @jafet-chaves, I just updated my class with your SIMO GList approach to handle the output caps  :slightly_smiling_face:.

tests/check/gsttiovxmultiscaler.c
dchvs: It's the unit test to the module initialization.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,458 @@
...
+ *
+ * FIXME:describe the real formats here.
+ */
+static GstStaticPadTemplate sink_template = GST_STATIC_PAD_TEMPLATE ("sink",
+    GST_PAD_SINK,
+    GST_PAD_ALWAYS,
+    GST_STATIC_CAPS (TIOVX_MULTI_SCALER_STATIC_CAPS_SINK)
+    );
+
+static GstStaticPadTemplate src_template = GST_STATIC_PAD_TEMPLATE ("src",
```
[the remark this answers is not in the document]
dchvs: Alright.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,458 @@
...
+      g_param_spec_int ("num-outputs", "Multi Scaler number of outputs",
+      "Multi Scaler number of outputs", DEFAULT_PROP_MIN_NUM_OUTPUTS,
+      DEFAULT_PROP_MAX_NUM_OUTPUTS, DEFAULT_PROP_NUM_OUTPUTS,
+      (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS));
+
+  g_object_class_install_properties (gobject_class, N_PROPERTIES,
+      obj_properties);
+
+  GST_DEBUG_CATEGORY_INIT (gst_tiovx_multi_scaler_debug,
+      "tiovxmultiscaler", 0, "debug category for the tiovxmultiscaler element");
```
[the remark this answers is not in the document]
dchvs: As agreed, let's keep it as TI wants it, tiovxmultiscale.

tests/check/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,189 @@
...
+
+  gst_init (NULL, NULL);
+  element = gst_element_factory_make ("tiovxmultiscaler", "tiovxmultiscaler");
+
+  simo_class = GST_TIOVX_SIMO_GET_CLASS (element);
+  simo = GST_TIOVX_SIMO (element);
+  in_pool_size = g_random_int_range (MIN_POOL_SIZE, MAX_POOL_SIZE);
+  out_pool_sizes = g_hash_table_new (NULL, NULL);
+
+  src_caps_list = g_list_alloc ();
```
jafet-chaves, line 129: This creates a bug, the first GList you are leaving it empty, I recommend to use g_list_alloc or g_list_append instead always to expand the list and underneath g_list_alloc is called

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+      GST_DEBUG_FUNCPTR (gst_tiovx_multi_scaler_configure_module);
+
+  tiovx_simo_class->create_graph =
+      GST_DEBUG_FUNCPTR (gst_tiovx_multi_scaler_create_graph);
+
+  tiovx_simo_class->deinit_module =
+      GST_DEBUG_FUNCPTR (gst_tiovx_multi_scaler_deinit_module);
+
+  obj_properties[PROP_DEFAULT_TARGET] =
+      g_param_spec_string ("default-target", "TIVX default target",
```
michaelgruner, line 209: We want this to be a `g_param_spec_enum` not a string. Please tackle this in a new pr

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+    GST_STATIC_CAPS (TIOVX_MULTI_SCALER_STATIC_CAPS_SINK)
+    );
+
+static GstStaticPadTemplate src_template = GST_STATIC_PAD_TEMPLATE ("src",
+    GST_PAD_SRC,
+    GST_PAD_ALWAYS,
+    GST_STATIC_CAPS (TIOVX_MULTI_SCALER_STATIC_CAPS_SRC)
+    );
+
+static ScalerObj _scaler_obj;
```
michaelgruner, line 134: No, we can never have this global. This will break as soon as we have two instances of the multiscaler. they will both attempt to use this global instance.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+    const GValue * value, GParamSpec * pspec)
+{
+  GstTIOVXMultiScaler *self = GST_TIOVX_MULTI_SCALER (object);
+
+  GST_LOG_OBJECT (self, "set_property");
+
+  GST_OBJECT_LOCK (self);
+  switch (prop_id) {
+    case PROP_DEFAULT_TARGET:
+      self->default_target = g_value_get_string (value);
```
michaelgruner, line 250: for strings you always use `g_value_dup_string`. This means that you need to free the previous string.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+      "tiovxmultiscaler", 0, "debug category for the tiovxmultiscaler element");
+}
+
+/* Initialize the new element
+ * Initialize instance structure
+ */
+static void
+gst_tiovx_multi_scaler_init (GstTIOVXMultiScaler * self)
+{
+  self->scaler_obj = &_scaler_obj;
```
michaelgruner, line 232: This cannot be a global instance.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+
+  self->default_target = DEFAULT_PROP_TARGET;
+  self->num_channels = DEFAULT_PROP_NUM_CHANNELS;
+  self->num_outputs = DEFAULT_NUM_OUTPUTS;
```
michaelgruner, line 236: Initialize graph

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+
+/* Initialize the new element
+ * Initialize instance structure
+ */
+static void
+gst_tiovx_multi_scaler_init (GstTIOVXMultiScaler * self)
+{
+  self->scaler_obj = &_scaler_obj;
+
+  self->default_target = DEFAULT_PROP_TARGET;
```
michaelgruner, line 234: In the `set_property` you'll need to use `g_value_dup_string` which forces you do free the last string. As such, make a copy here (using g_strdup) so that you can free it without problems there.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+
+  multiscaler->num_outputs = (vx_int32) self->num_outputs;
+  gst_video_info_from_caps (&in_info, sink_caps);
+  multiscaler->input.width = (vx_int32) GST_VIDEO_INFO_WIDTH ((&in_info));
+  multiscaler->input.height = (vx_int32) GST_VIDEO_INFO_HEIGHT ((&in_info));
+  multiscaler->color_format =
+      gst_tiovx_utils_map_gst_video_format_to_vx_format (in_info.finfo->format);
+
+  multiscaler->method = VX_INTERPOLATION_BILINEAR;
+  multiscaler->input.bufq_depth = (vx_int32) in_pool_size;
```
michaelgruner, line 327: Remove uneeded casts.

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+
+  /* Initialize the output parameters */
+  for (i = 0; i < multiscaler->num_outputs; i++) {
+    src_caps = g_list_next (src_caps_list)->data;
```
michaelgruner, line 331: This will always return the same caps. You are not iterating. You need to do something like:
```c
GstCaps *caps = (GstCaps *)src_caps_list->data;
src_caps_list = g_list_next (src_caps_list);
```

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+
+  self = GST_TIOVX_MULTI_SCALER (simo);
+
+  status = vxVerifyGraph (self->graph);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self,
+        "Module configure verify graph failed with error: %d", status);
+    ret = FALSE;
+    goto out;
+  }
```
michaelgruner, line 384: Please confirm with @jafet-chaves if this is done in the parent class

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+  status =
+      app_create_graph_scaler (context, graph, multiscaler, NULL,
+      default_target);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Create graph failed with error: %d", status);
+    ret = FALSE;
+    goto out;
+  }
+
+  self->graph = graph;
```
michaelgruner, line 435: Why do we need to save this? we shouldn't

ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -0,0 +1,472 @@
...
+    ret = FALSE;
+    goto out;
+  }
+
+  status = app_deinit_scaler (multiscaler);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Module deinit failed with error: %d", status);
+    ret = FALSE;
+    goto out;
+  }
```
michaelgruner, line 468: Is this order okay? we delete first and then deinit? sounds strange to me.

michaelgruner: Also, null out everything you free.

ext/tiovx/gsttiovxmultiscaler.h
```diff
@@ -0,0 +1,83 @@
...
+
+G_BEGIN_DECLS
+#define GST_TYPE_GST_TIOVX_MULTI_SCALER (gst_tiovx_multi_scaler_get_type())
+G_DECLARE_FINAL_TYPE(GstTIOVXMultiScaler, gst_tiovx_multi_scaler, GST,
+                     TIOVX_MULTI_SCALER, GstTIOVXSimo)
+
+G_END_DECLS
+
+#define MIN_POOL_SIZE 2
+#define MAX_POOL_SIZE 16
```
michaelgruner, line 80: Where did this value come from?

ext/tiovx/gsttiovxmultiscaler.h
```diff
@@ -0,0 +1,83 @@
...
+#include "gst-libs/gst/tiovx/gsttiovxsimo.h"
+
+G_BEGIN_DECLS
+#define GST_TYPE_GST_TIOVX_MULTI_SCALER (gst_tiovx_multi_scaler_get_type())
+G_DECLARE_FINAL_TYPE(GstTIOVXMultiScaler, gst_tiovx_multi_scaler, GST,
+                     TIOVX_MULTI_SCALER, GstTIOVXSimo)
+
+G_END_DECLS
+
+#define MIN_POOL_SIZE 2
```
michaelgruner, line 79: Move to the source file.

michaelgruner (changes requested):

ext/tiovx/gsttiovxmultiscaler.c
dchvs: We keep it to verify the graph later on.

ext/tiovx/gsttiovxmultiscaler.h
dchvs: It corresponds to the max. `bufq_depth` the modules image object supports.

michaelgruner (approved):

PR #34, opened 2021-08-04
Fix/modules phase 1.2 rel1

shyam-j:
Updates to color-convert and multi-scaler modules, mostly changes in API names.
_create module functions don't require context handle. Some opportunity to clean the GST wrapper function.


shyam-j:
Incorrect merge request

PR #44, opened 2021-08-10
Fix/colorconvert handling of non tiovx buffers from upstream

jafet-chaves:


gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -326,6 +337,43 @@ gst_tiovx_siso_set_caps (GstBaseTransform * trans, GstCaps * incaps,
   return ret;
 }
 
+static GstBuffer *
+gst_tiovx_siso_buffer_copy (GstTIOVXSiso * self, GstTIOVXBufferPool * pool,
```
emmadrigal, line 341: We already have `simo_buffer_copy` right? lets factor out this logic to `utils`

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -403,6 +473,10 @@ gst_tiovx_siso_transform (GstBaseTransform * trans, GstBuffer * inbuf,
     goto free;
   }
 
+  if (free_inbuf) {
+    gst_buffer_unref (inbuf);
```
emmadrigal, line 477: After freeing we need to null out the variable:
```
inbuf = NULL;
```

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -447,8 +522,9 @@ gst_tiovx_siso_decide_allocation (GstBaseTransform * trans, GstQuery * query)
     /* Create our own pool if a TIOVX was not found.
        We use output vx_reference to decide a pool to use downstream. */
     ret =
-        gst_tiovx_add_new_pool (GST_CAT_DEFAULT, query, priv->out_pool_size,
-        priv->output, &priv->out_info);
+        gst_tiovx_siso_add_new_pool (self, query, priv->out_pool_size,
+        priv->output, &priv->out_info, &pool);
+    gst_object_unref (pool);
```
emmadrigal, line 527: null-out `pool` after this

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -460,8 +536,10 @@ gst_tiovx_siso_propose_allocation (GstBaseTransform * trans,
 {
   GstTIOVXSiso *self = GST_TIOVX_SISO (trans);
   GstTIOVXSisoPrivate *priv = gst_tiovx_siso_get_instance_private (self);
+  GstBufferPool *pool = NULL;
   gboolean ret = TRUE;
 
+
```
emmadrigal, line 542: Is this line necessary?

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -474,10 +552,12 @@ gst_tiovx_siso_propose_allocation (GstBaseTransform * trans,
...
   /* We use input vx_reference to propose a pool upstream */
-
   ret =
-      gst_tiovx_add_new_pool (GST_CAT_DEFAULT, query, priv->in_pool_size,
-      priv->input, &priv->in_info);
+      gst_tiovx_siso_add_new_pool (self, query, priv->in_pool_size,
+      priv->input, &priv->in_info, &pool);
+  priv->sink_buffer_pool = GST_TIOVX_BUFFER_POOL (pool);
+  gst_object_ref (priv->sink_buffer_pool);
+  gst_object_unref (pool);
```
emmadrigal, line 560: I'm not sure if the order of ref-unref here is correct.
The last two lines here actually have opposite effects on the same variable.

I think it should be more like:

```c
// Unref the old stored pool
if (NULL != priv->sink_buffer_pool)
  gst_object_unref (priv->sink_buffer_pool);
// Assign the new pool to the internal value
priv->sink_buffer_pool = GST_TIOVX_BUFFER_POOL (pool);
// Ref back the new pool
gst_object_ref(pool);
```
However, if `add_new_pool` actually transfers the new pool to you `ref`ing at the end is uncessary

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -807,6 +887,54 @@ gst_tiovx_siso_process_graph (GstTIOVXSiso * self)
...
+    GstBufferPool ** pool)
+{
+  GstCaps *caps = NULL;
+  GstStructure *config = NULL;
+  gsize size = 0;
+
+  g_return_val_if_fail (self, FALSE);
+  g_return_val_if_fail (query, FALSE);
+  g_return_val_if_fail (info, FALSE);
+  g_return_val_if_fail (exemplar, FALSE);
```
emmadrigal, line 902: `pool` pointer guard

emmadrigal (changes requested):

gst-libs/gst/tiovx/gsttiovxutils.c
```diff
@@ -276,3 +278,40 @@ gst_tiovx_add_new_pool (GstDebugCategory * category, GstQuery * query,
 
   return TRUE;
 }
+
+GstBuffer *
+gst_tiovx_buffer_copy (GstObject * self, GstTIOVXBufferPool * pool,
```
emmadrigal, line 283: Since this function isn't actually located in a file where `GST_DEFAULT_CAT` is defined and can be called from multiple elements, I've found the best is to receive `GstDebugCategory * category` instead of `GstObject`

And replacing `GST_ERROR_OBJECT (self, ...` with `GST_CAT_ERROR(category` or `GST_CAT_ERROR_OBJECT(category, self` if you still want the error to belong to an object.

Not doing this will cause the debug to not be reported with the correct type and us not being able to filter them with for example: `GST_DEBUG=*tiovx*:6`

gst-libs/gst/tiovx/gsttiovxutils.h
```diff
@@ -120,4 +122,17 @@ gboolean
 gst_tiovx_add_new_pool (GstDebugCategory * category, GstQuery * query,
     guint num_buffers, vx_reference * exemplar, GstVideoInfo * info);
 
+/**
+ * gst_tiovx_buffer_copy:
+ * @self: Object using this function
+ * @pool: Pool from which to extract output buffer to copy into
+ * @in_buffer: Input buffer that will be copied
```
emmadrigal, line 129: Add: `This function doesn't take ownership of in_buffer`

emmadrigal (changes requested):

gst-libs/gst/tiovx/gsttiovxutils.c
```diff
@@ -276,3 +278,40 @@ gst_tiovx_add_new_pool (GstDebugCategory * category, GstQuery * query,
 
   return TRUE;
 }
+
+GstBuffer *
+gst_tiovx_buffer_copy (GstDebugCategory * category, GstTIOVXBufferPool * pool,
```
michaelgruner, line 283: No need to ask for the `GstTIOVXBufferPool`. We can ask for a buffer pool so we don't need an extra include.

michaelgruner (approved):

PR #60, opened 2021-08-20
Fix case when the upstream element does not manage a buffer pool

jafet-chaves:


gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -376,6 +379,49 @@ gst_tiovx_siso_transform (GstBaseTransform * trans, GstBuffer * inbuf,
...
+    if (!gst_tiovx_configure_pool (GST_CAT_DEFAULT, pool, priv->input,
+            priv->in_caps, size, priv->in_pool_size)) {
+      GST_ERROR_OBJECT (self, "Unable to configure pool in transform function");
+      return FALSE;
+    }
+
+    /* Unref the old stored pool */
+    if (NULL != priv->sink_buffer_pool) {
+      gst_object_unref (priv->sink_buffer_pool);
+    }
```
emmadrigal, line 412: Is this even reachable since you are checking for:
```
NULL == priv->sink_buffer_pool
```
at the top

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -376,6 +379,49 @@ gst_tiovx_siso_transform (GstBaseTransform * trans, GstBuffer * inbuf,
...
+    /* Assign the new pool to the internal value */
+    priv->sink_buffer_pool = GST_BUFFER_POOL (pool);
+
+    inbuf =
+        gst_tiovx_buffer_copy (GST_CAT_DEFAULT,
+        GST_BUFFER_POOL (priv->sink_buffer_pool), inbuf);
+    if (!inbuf) {
+      GST_ERROR_OBJECT (self, "Failure when copying input buffer from pool");
+      goto exit;
```
emmadrigal, line 421: Let's put the `  if (NULL == priv->sink_buffer_pool) {` logic before the `/* Was the pool accepted by the upstream element? */`.
This way we avoid doing this copy logic twice

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -376,6 +379,49 @@ gst_tiovx_siso_transform (GstBaseTransform * trans, GstBuffer * inbuf,
       free_inbuf = TRUE;
     }
   }
+  /* Propose allocation did not happen, there is not upstream pool therefore
```
emmadrigal, line 382: Typo: `there is no upstream pool`

emmadrigal (changes requested):

rrcarlosrodriguez (approved):

gst-libs/gst/tiovx/gsttiovxsiso.c
```diff
@@ -352,7 +352,41 @@ gst_tiovx_siso_transform (GstBaseTransform * trans, GstBuffer * inbuf,
...
+    pool = gst_tiovx_create_new_pool (GST_CAT_DEFAULT, priv->input);
+    if (NULL == pool) {
+      GST_ERROR_OBJECT (self,
+          "Failed to create new pool in transform function");
+      return FALSE;
+    }
+
+    if (!gst_tiovx_configure_pool (GST_CAT_DEFAULT, pool, priv->input,
+            priv->in_caps, size, priv->in_pool_size)) {
+      GST_ERROR_OBJECT (self, "Unable to configure pool in transform function");
```
emmadrigal, line 381: You need to free `pool` in this failure case

emmadrigal (approved):

gst-libs/gst/tiovx/gsttiovxsiso.c
jafet-chaves: The pool is already freed in case of failure internally in `gst_tiovx_configure_pool` : https://github.com/TexasInstruments/edgeai-gst-plugins/blob/fix/no-upstream-pool-case/gst-libs/gst/tiovx/gsttiovxutils.c#L261

Issue #61, opened 2021-08-23
Output is not correct whenever not tiovx buffers are processed by tiovx plugins


ravi-rahul:
GST_DEBUG=tiovx*:4 gst-launch-1.0 multifilesrc location=/opt/i420/%010u.yuv index=100 loop=true ! video/x-raw, format=I420, width=1280, height=720, framerate=30/1 ! tiovxcolorconvert !  video/x-raw, format=NV12 ! kmssink
 
 output is not correct for this pipeline
 similar behavior whenever non tiovx buffers are processed

ravi-rahul:
the issue can also be reproduced in the application using below steps

1. sync the edge_ai_apps git repo to get the updated video file
2. Modify the config file configs/image_classification.yaml to set the input as video like below
         flows:
             flow0:
                 input: input1 (change input0 to input1)
                 
 3. run the application
      cd apps_python
      ./app_edgeai.py ../configs/image_classification.yaml

ravi-rahul:
first run the demo with camera input
after that if we run demo with file input we can see the traces of camera frame in the output

ravi-rahul:
Hi @emmadrigal 

I tested the application with fix/memory-copy branch
the issue is not coming, getting correct output

rrcarlosrodriguez:
That's great @ravi-rahul ! We will proceed with the merge into develop

Feel free to close this issue whenever you want

PR #62, opened 2021-08-23
Make a manual copy to TIs memory

emmadrigal:
Using gst_buffer_copy will create a new memory appended to our buffer,
we are now manually copying the memory from the input buffer to ensure
valid data in the meta.

### DIFF
--- a/gst-libs/gst/tiovx/gsttiovxutils.c
+++ b/gst-libs/gst/tiovx/gsttiovxutils.c
@@ -65,6 +65,7 @@
 
 #include <TI/j7.h>
 
+#include "gsttiovxallocator.h"
 #include "gsttiovxbufferpool.h"
 #include "gsttiovxmeta.h"
 #include "gsttiovxtensorbufferpool.h"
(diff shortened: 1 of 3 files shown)


emmadrigal:
Fixes #61

gst-libs/gst/tiovx/gsttiovxutils.c
```diff
@@ -394,13 +398,26 @@ gst_tiovx_buffer_copy (GstDebugCategory * category, GstBufferPool * pool,
...
+        "Error copying flags from in buffer: %" GST_PTR_FORMAT
+        " to out buffer: %" GST_PTR_FORMAT, in_buffer, out_buffer);
     gst_buffer_unref (out_buffer);
     out_buffer = NULL;
     goto out;
   }
 
+  gst_buffer_map (in_buffer, &in_info, GST_MAP_READ);
+
+  memory = gst_buffer_get_memory (out_buffer, 0);
```
rrcarlosrodriguez, line 410: Please check if this has to be freed, if so, there is a huge memory leak

rrcarlosrodriguez (changes requested):

PR #69, opened 2021-08-30
Compare caps before modules reinitialization

rrcarlosrodriguez:
If caps are equal, depending on subclass analysis, the modules
are deinit/init in order to accept new caps.


michaelgruner (approved):

emmadrigal (approved):

jafet-chaves (approved):

gst-libs/gst/tiovx/gsttiovxsiso.h
```diff
@@ -104,6 +104,8 @@ G_DECLARE_DERIVABLE_TYPE (GstTIOVXSiso, gst_tiovx_siso, GST,
  *                      vx_reference memory allocated.
  * @deinit_module:      Required. Subclasses must override to deinit
  *                      the element-specific module.
+ * @compare_caps:       Optional. Subclasses must override to compare
+ *                      caps based on pad direction
```
jafet-chaves, line 108: Extend comment to provide insight about the expected return value in the base class. Your current implementation expects  TRUE if caps are the same.

PR #71, opened 2021-08-31
Use separate graph and node idx in MISO

emmadrigal:


gst-libs/gst/tiovx/gsttiovxmiso.c
```diff
@@ -83,7 +86,8 @@ typedef struct _GstTIOVXMisoPadPrivate
 
   guint pool_size;
   vx_reference *exemplar;
-  gint param_id;
+  gint graph_param_id;
```
michaelgruner, line 89: OpenVX handles these as unsigned integers, will this be a problem?

michaelgruner (approved):

rrcarlosrodriguez (approved):

Issue #75, opened 2021-09-02
MEM: ERROR: when multitiple tiovx plugins are used in the same pipeline


ravi-rahul:
Pipeline:

GST_DEBUG=tiovx*:4 gst-launch-1.0 videotestsrc is-live=true ! video/x-raw, width=1280, height=720, format=I420 ! tiovxcolorconvert ! video/x-raw, format=NV12 ! tiovxmultiscaler ! video/x-raw, width=454, height=256 ! tiovxcolorconvert ! video/x-raw, format=RGB ! fakesink

Log:

^Chandling interrupt.
Interrupt: Stopping pipeline ...
Execution ended after 0:00:01.359334270
Setting pipeline to NULL ...
0:00:01.437071045  3280     0x11040160 INFO       tiovxcolorconvert gsttiovxcolorconvert.c:613:gst_tiovx_color_convert_deinit_module:<tiovxcolorconvert1> Deinit module
0:00:01.437684100  3280     0x11040160 INFO       tiovxcolorconvert gsttiovxcolorconvert.c:613:gst_tiovx_color_convert_deinit_module:<tiovxcolorconvert0> Deinit module
**MEM: ERROR: Failed to free memory at virt addr = 0xffffa80b0000 !!!**
Freeing pipeline ...
**MEM: ERROR: Failed to free memory at virt addr = 0xffffa82a0000 !!!**

shyam-j:
The removed the VX_ZONE_ERRORS but introduces VX_ZONE_WARNING as below. Basically there are some hanging references which are collected during vxReleaseContext. Interestingly I see that this occurs during alternate runs, 
Run1 - no error, no warning,
Run2 - no error, warning present
Run3 - no error, no warning,
Run4 - no error, warning present

Here is an updated pipeline I tried,
[...]

PR #85, opened 2021-09-06
Hotfix: Attempt to push buffer to all pads after error

emmadrigal:
This fixes a memory leak, where after adquiring all input buffers, if
one failed to push not all inputs were released. This fix attempts to
push to all pads, which correctly release all input buffers and allows
the corresponding bufferpool to be released.


emmadrigal:
This fixes #75

jafet-chaves (approved):

rrcarlosrodriguez (approved):

PR #88, opened 2021-09-07
Fix/multiscaler renegotiation on different caps

jafet-chaves:


ext/tiovx/gsttiovxmultiscaler.c
```diff
@@ -865,3 +872,35 @@ target_id_to_target_name (gint target_id)
...
+
+static gboolean
+gst_tiovx_multi_scaler_compare_caps (GstTIOVXSimo * self, GstCaps * caps1,
+    GstCaps * caps2, GstPadDirection direction)
+{
+  GstVideoInfo video_info1;
+  GstVideoInfo video_info2;
+  gboolean ret = FALSE;
+
+  g_return_val_if_fail (caps1, FALSE);
```
emmadrigal, line 884: self pointer guard

emmadrigal (approved):

Issue #84, opened 2021-09-03
MISO without buffers timestamps


jafet-chaves:
Pipeline to reproduce
```
GST_DEBUG=2,GST_TRACER:7 GST_TRACERS=leaks G_DEBUG=fatal-warnings gst-launch-1.0 \
videotestsrc num-buffers=5 ! "video/x-raw,format=RGB,width=1280,height=720" ! \
tiovxcolorconvert ! "video/x-raw,format=NV12,width=1280,height=720" ! \
tiovxmultiscaler ! "video/x-raw,format=NV12,width=640,height=480" ! queue ! mosaic.sink_0 \
videotestsrc num-buffers=5 ! "video/x-raw,format=RGB,width=1280,height=720" ! \
tiovxcolorconvert ! "video/x-raw,format=NV12,width=1280,height=720" ! \
tiovxmultiscaler ! "video/x-raw,format=NV12,width=640,height=480" ! queue ! mosaic.sink_1 \
videotestsrc num-buffers=5 ! "video/x-raw,format=RGB,width=1280,height=720" ! \
tiovxcolorconvert ! "video/x-raw,format=NV12,width=1280,height=720" ! \
tiovxmultiscaler ! "video/x-raw,format=NV12,width=640,height=480" ! queue ! mosaic.sink_2 \
videotestsrc num-buffers=5 ! "video/x-raw,format=RGB,width=1280,height=720" ! \
tiovxcolorconvert ! "video/x-raw,format=NV12,width=1280,height=720" ! \
tiovxmultiscaler ! "video/x-raw,format=NV12,width=640,height=480" ! queue ! mosaic.sink_3 \
tiovxmosaic name=mosaic \
sink_0::startx=0   sink_0::starty=0   \
sink_1::startx=640 sink_1::starty=0   \
sink_2::startx=0 sink_2::starty=480   \
sink_3::startx=640 sink_3::starty=480  \
! \
"video/x-raw,format=NV12" ! \
fpsdisplaysink video-sink=fakesink -v -e
```

PR #96, opened 2021-09-09
Fix/miso add outbuf timestamp info

jafet-chaves:
This is a fix proposal for #84

### DIFF
--- a/gst-libs/gst/tiovx/gsttiovxmiso.c
+++ b/gst-libs/gst/tiovx/gsttiovxmiso.c
@@ -74,6 +74,8 @@
 
 #define DEFAULT_POOL_SIZE MIN_POOL_SIZE
 #define MAX_NUMBER_OF_PLANES 4
+#define GST_BUFFER_OFFSET_FIXED_VALUE -1
+#define GST_BUFFER_OFFSET_END_FIXED_VALUE -1
 
 GST_DEBUG_CATEGORY_STATIC (gst_tiovx_miso_debug_category);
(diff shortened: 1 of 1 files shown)


gst-libs/gst/tiovx/gsttiovxmiso.c
```diff
@@ -508,6 +510,9 @@ gst_tiovx_miso_aggregate (GstAggregator * agg, gboolean timeout)
   GstBuffer *outbuf = NULL;
   GstFlowReturn ret = GST_FLOW_ERROR;
   GList *l = NULL;
+  GstClockTime pts = G_MAXUINT64;
```
michaelgruner, line 513: GST_CLOCK_TIME_NONE

emmadrigal (approved):

gst-libs/gst/tiovx/gsttiovxmiso.c
```diff
@@ -531,9 +536,27 @@ gst_tiovx_miso_aggregate (GstAggregator * agg, gboolean timeout)
...
+    /* Find the smallest timestamp and the largest duration */
+    if (tmp_pts < pts) {
+        pts = tmp_pts;
+    }
+    if (tmp_dts < dts) {
+        dts = tmp_dts;
+    }
+    if (tmp_duration > duration) {
+        duration = tmp_duration;
+    }
```
michaelgruner, line 558: Thinking this through, I don't feel we should do this. I would generate a theoretical duration based on our framerate.

michaelgruner (changes requested):

Issue #68, opened 2021-08-27
Small memory leak with colorconvert


emmadrigal:
The following pipeline:
```
GST_DEBUG=multifilesrc*:LOG,tiovx*:LOG gdb --args gst-launch-1.0 multifilesrc location=/opt/edge_ai_apps/data/images/%04d.jpg loop=true caps="image/jpeg, framerate=1/1" !  jpegdec ! videoscale ! "video/x-raw, width=1280, height=720, pixel-aspect-ratio=(fraction)1/1" ! tiovxcolorconvert in-pool-size=4 out-pool-size=4 ! video/x-raw, format=NV12 ! fakesink
```
Has a small memory leak, during testing after ~45min the memory raises from 2.2% to 5.6%.

This pipeline is however also causing renegotiation with causes buffer pools and allocators to also be deleted and recreated so errors might be there that do not affect pipelines with changing caps

PR #97, opened 2021-09-10
Fix gstcaps leaks in gsttiovxsiso

m-herrera:


emmadrigal (approved):

m-herrera:
This solves colorconvert memory leaks reported here: #68

rrcarlosrodriguez (approved):

PR #100, opened 2021-09-17
Feature/extend colorconvert unit tests

rrcarlosrodriguez:


tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,43 +89,224 @@ static const gchar *gst_sink_formats[] = {
...
+  {"NV21", "UYVY", "Y444"},
+  {"NV21", "UYVY", "YUY2"}
+};
+
+static const gchar *pipelines_caps_negotiation_fail[] = {
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,width=320,height=240 ! tiovxcolorconvert ! video/x-raw,width=640,height=480 ! fakesink async=false",
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,width=320 ! tiovxcolorconvert ! video/x-raw,width=640 ! fakesink async=false",
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,height=240 ! tiovxcolorconvert ! video/x-raw,height=480 ! fakesink async=false",
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,width=[320, 640],height=240 ! tiovxcolorconvert ! video/x-raw,height=480 ! fakesink async=false",
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,width=320,height=[240, 480] ! tiovxcolorconvert ! video/x-raw,width=640 ! fakesink async=false",
```
michaelgruner, line 117: We must stop using async=false without a real reason to do so. This is harmful for our tests because we report ourselves in  a playing state before a buffer actually reaches the sink. We might lose some errors.

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,43 +89,224 @@ static const gchar *gst_sink_formats[] = {
...
+      buffer_counter);
+
+  caps_320x240 =
+      gst_caps_from_string ("video/x-raw,width=320,height=240,format=NV12");
+  caps_640x480 =
+      gst_caps_from_string ("video/x-raw,width=640,height=480,format=NV12");
+
+  while (renegotiation_attempts < NUM_RENEGOTIATION_ATTEMPTS) {
+    g_mutex_lock (&buffer_counter->mutex);
+    g_cond_wait (&buffer_counter->cond, &buffer_counter->mutex);
```
michaelgruner, line 272: I believe we need to use a timed wait to avoid a deadlock in case of an error.

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,43 +89,224 @@ static const gchar *gst_sink_formats[] = {
...
+  const gchar *desc =
+      "videotestsrc is-live=true ! capsfilter name=caps caps=video/x-raw,width=320,height=240,format=NV12 ! tiovxcolorconvert ! video/x-raw,format=RGB ! fakesink name=fakesink async=false signal-handoffs=true";
+  GstElement *pipeline = NULL;
+  GstElement *caps = NULL;
+  GstElement *fakesink = NULL;
+  GstCaps *caps_320x240 = NULL;
+  GstCaps *caps_640x480 = NULL;
+  guint renegotiation_attempts = 0;
+  gboolean caps_switch = TRUE;
+  BufferCounter *buffer_counter = g_malloc0 (sizeof (BufferCounter));
```
michaelgruner, line 241: No need for dynamic memory here.

tests/check/test_utils.h
```diff
@@ -111,46 +123,73 @@ test_create_pipeline_fail (const gchar * pipe_desc)
...
+  pipeline = test_create_pipeline (pipe_desc);
+
+  for (i = 0; i < num_state_changes; i++) {
+    fail_unless_equals_int (gst_element_set_state (pipeline, GST_STATE_PLAYING),
+        state_change[0]);
+    fail_unless_equals_int (gst_element_get_state (pipeline, NULL, NULL,
+            GST_CLOCK_TIME_NONE), state_change[1]);
+    fail_unless_equals_int (gst_element_set_state (pipeline, GST_STATE_NULL),
+        state_change[2]);
+  }
```
michaelgruner, line 141: This means nothing. We really need to fix this. We are not capturing anything here, specially if we use async=false. I bet not even caps get negotiated here. We need to wait and validate we actually get to the playing state. Also, errors creating the graph are reported to the bus, we won't catch them here. This doesn't test that much really.

tests/check/test_utils.h
```diff
@@ -111,46 +123,73 @@ test_create_pipeline_fail (const gchar * pipe_desc)
...
+
+  if (buffer_counter->buffer_limit < buffer_counter->buffer_counter) {
+    g_mutex_lock (&buffer_counter->mutex);
+    g_cond_signal (&buffer_counter->cond);
+    g_mutex_unlock (&buffer_counter->mutex);
+    buffer_counter->buffer_counter = 0;
+    return;
+  }
+  buffer_counter->buffer_counter++;
+}
```
michaelgruner, line 192: Also, don't place definition on headers. There's no point in compiling the same code over and over again. Always create a compilation unit for this. Don't change it, I already absorbed this in my branch.

michaelgruner (changes requested):

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,31 +89,206 @@ static const gchar *gst_sink_formats[] = {
...
+  "videotestsrc is-live=true num-buffers=5 ! video/x-raw,width=[320, 640],height=[240, 480] ! tiovxcolorconvert ! video/x-raw,width=[320, 640],height=[240, 480] ! fakesink ",
+  NULL,
 };
 
-GST_START_TEST (test_state_change)
+GST_START_TEST (test_state_change_success)
+{
+  gchar pipeline[MAX_PIPELINE_SIZE] = "";
+  gint sink_format;
+  gint src_format;
```
emmadrigal, line 144: Initialize to 0

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,31 +89,206 @@ static const gchar *gst_sink_formats[] = {
...
+  }
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_pad_creation_fail)
 {
   gchar pipeline[MAX_PIPELINE_SIZE] = "";
   gint sink_format;
   gint src_format;
```
emmadrigal, line 172: Missing initialization

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,31 +89,206 @@ static const gchar *gst_sink_formats[] = {
...
+  /* Connect to fakesink signal to count number of buffers */
+  g_signal_connect (fakesink, "handoff", G_CALLBACK (buffer_counter_func),
+      &buffer_counter);
+
+  caps_320x240 =
+      gst_caps_from_string ("video/x-raw,width=320,height=240,format=NV12");
+  caps_640x480 =
+      gst_caps_from_string ("video/x-raw,width=640,height=480,format=NV12");
+
+  while (renegotiation_attempts < NUM_RENEGOTIATION_ATTEMPTS) {
```
emmadrigal, line 269: A `for` loop is better in this case

tests/check/gsttiovxcolorconvert.c
```diff
@@ -84,31 +89,206 @@ static const gchar *gst_sink_formats[] = {
...
+  const gchar *desc =
+      "videotestsrc is-live=true ! capsfilter name=caps caps=video/x-raw,width=320,height=240,format=NV12 ! tiovxcolorconvert ! video/x-raw,format=RGB ! fakesink name=fakesink signal-handoffs=true";
+  GstElement *pipeline = NULL;
+  GstElement *caps = NULL;
+  GstElement *fakesink = NULL;
+  GstCaps *caps_320x240 = NULL;
+  GstCaps *caps_640x480 = NULL;
+  guint renegotiation_attempts = 0;
+  gboolean caps_switch = TRUE;
+  BufferCounter buffer_counter;
```
emmadrigal, line 240: ```  BufferCounter buffer_counter = {};```

emmadrigal (changes requested):

Issue #86, opened 2021-09-06
Mosaic: Support background image


rrcarlosrodriguez:

jafet-chaves:
Solved by #102

PR #102, opened 2021-09-21
Feature/add background image to mosaic

emmadrigal:


emmadrigal:
Add support for #86

gst-libs/gst/tiovx/gsttiovxmiso.h
```diff
@@ -164,8 +164,11 @@ struct _GstTIOVXMisoPadClass
  * @graph_param_id: Parameter id that will be used to enqueue this parameter to the Vx Graph
  * @node_param_id: Parameter id that will be used to configure the node
  *
- * Sets an exemplar an a param id to the pad, these will be used for future
- * configuration of the given pad.
+ * Sets an exemplar an a param id to the pad, these will be used for future configuration of the
+ * given pad.
+ *
+ * @graph_param_id and @node_param_id can be both set to -1 in order ot not use these for node
```
jafet-chaves, line 170: set to -1 in order or not use?

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -853,21 +906,83 @@ gst_tiovx_mosaic_fixate_caps (GstTIOVXMiso * self,
...
+          GST_PTR_FORMAT " in pad: %" GST_PTR_FORMAT, caps, background_pad);
+      gst_caps_unref (caps);
+      goto out;
+    }
+    gst_caps_unref (caps);
+
+    background_width = GST_VIDEO_INFO_WIDTH (&video_info);
+    background_height = GST_VIDEO_INFO_HEIGHT (&video_info);
+
+    if ((best_width <= background_width) && (best_height <= background_height)) {
```
michaelgruner, line 928: This should be an || I believe

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -853,21 +906,83 @@ gst_tiovx_mosaic_fixate_caps (GstTIOVXMiso * self,
...
+        (G_VALUE_TYPE (candidate_height_value) != G_TYPE_INT)) {
+      GST_ERROR_OBJECT (self,
+          "Width and height couldn't be fixated to : %" GST_PTR_FORMAT,
+          candidate_output_structure);
+    } else if ((g_value_get_int (candidate_width_value) != best_width)
+        || (g_value_get_int (candidate_height_value) != best_height)) {
+      GST_ERROR_OBJECT (self,
+          "Could not fixate: (%d, %d) to current source caps: %" GST_PTR_FORMAT,
+          best_width, best_height, candidate_output_structure);
+    }
```
michaelgruner, line 964: Simplify this by simply calling `gst_structure_get_int`

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -853,21 +906,83 @@ gst_tiovx_mosaic_fixate_caps (GstTIOVXMiso * self,
...
+      goto out;
+    }
+
+    gst_structure_fixate_field_nearest_int (candidate_output_structure, "width",
+        best_width);
+    gst_structure_fixate_field_nearest_int (candidate_output_structure,
+        "height", best_height);
+    gst_structure_fixate_field_nearest_fraction (candidate_output_structure,
+        "framerate", best_fps_n, best_fps_d);
+  }
```
michaelgruner, line 985: Factor out this, the only difference is that at the end you validate that the size matches the background image. Also, what if the fixate resulted in a smaller size? is this a valid situation?

michaelgruner (changes requested):

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -638,9 +664,15 @@ gst_tiovx_mosaic_create_graph (GstTIOVXMiso * agg, vx_context context,
...
-      NULL, input_arr_user, target);
+  if (self->has_background) {
+    status =
+        tiovx_img_mosaic_module_create (graph, mosaic,
+        mosaic->background_image[0], input_arr_user, target);
+  } else {
+    status =
+        tiovx_img_mosaic_module_create (graph, mosaic,
+        NULL, input_arr_user, target);
+  }
```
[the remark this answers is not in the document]
emmadrigal: No, `background_image[0]` is always created by the mosaic_module

ext/tiovx/gsttiovxmosaic.c
emmadrigal: No, both best_width and height need to be smaller than the background size

ext/tiovx/gsttiovxmosaic.c
emmadrigal: The check for `gst_tiovx_mosaic_validate_candidate_dimension`

Ensures that the candidate_output_structure (which comes from the source caps) is at least big enough, otherwise it exists before trying to fixate

jafet-chaves (approved):

michaelgruner (approved):

PR #103, opened 2021-09-22
Feature/add dlpreproc tests

dchvs:


tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+
+enum
+{
+  /* Pipelines names */
+  PIPELINE_SIMPLE = 0,
+};
+
+/* Supported formats */
+#define TIOVXDLPREPROC_FORMATS_NUMBER 4
+static const gchar *tiovxdlpreproc_formats[TIOVXDLPREPROC_FORMATS_NUMBER + 1] = {
```
emmadrigal, line 84: I don't see the need for this variable, something like this its simpler:
```
static const gchar *tiovxdlpreproc_formats[] = {
```

and the following definition lower whenever you need it
```
#define ARRAYSIZE(x)  (sizeof(x) / sizeof((x)[0]))
```

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_state_transitions_fail)
+{
+  test_states_change (test_pipelines[PIPELINE_SIMPLE]);
+}
+
+GST_END_TEST;
```
emmadrigal, line 201: How is this test different from `test_state_transitions`?

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+#define TIOVXDLPREPROC_DIMENSIONS_RANGE_VALUE 1
+static const uint tiovxdlpreproc_width[TIOVXDLPREPROC_DIMENSIONS_RANGE_VALUE +
+    1] = { 1, 8192 };
+static const uint tiovxdlpreproc_height[TIOVXDLPREPROC_DIMENSIONS_RANGE_VALUE +
+    1] = { 1, 8192 };
+
+/* Supported framerate */
+#define TIOVXDLPREPROC_FRAMERATE_RANGE_VALUE 1
+static const uint tiovxdlpreproc_framerate[TIOVXDLPREPROC_DIMENSIONS_RANGE_VALUE
+    + 1] = { 1, 2147483647 };
```
emmadrigal, line 102: Replace by:
```
static const uint tiovxdlpreproc_width[] = { 1, 8192 };
static const uint tiovxdlpreproc_height[] = { 1, 8192 };

/* Supported framerate */
static const uint tiovxdlpreproc_framerate[] = { 1, 2147483647 };
```

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+
+/* Supported data-type */
+#define TIOVXDLPREPROC_DATA_TYPE_NUMBER 7
+static const gchar *tiovxdlpreproc_data_type[TIOVXDLPREPROC_DATA_TYPE_NUMBER +
+    1] = {
```
emmadrigal, line 107: See my comment about `ARRAYSIZE` above

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  "int32",
+  "uint32",
+  "float32",
+  NULL,
+};
+
+/* Supported channel-order */
+#define TIOVXDLPREPROC_CHANNEL_ORDER_NUMBER 2
+static const gchar
+    * tiovxdlpreproc_channel_order[TIOVXDLPREPROC_CHANNEL_ORDER_NUMBER + 1] = {
```
emmadrigal, line 121: See my comment about ARRAYSIZE above

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  "nchw",
+  "nhwc",
+  NULL,
+};
+
+/* Supported tensor-format */
+#define TIOVXDLPREPROC_TENSOR_FORMAT_NUMBER 2
+static const gchar
+    * tiovxdlpreproc_tensor_format[TIOVXDLPREPROC_TENSOR_FORMAT_NUMBER + 1] = {
```
emmadrigal, line 130: See my comment about ARRAYSIZE above

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  "rgb",
+  "bgr",
+  NULL,
+};
+
+/* Supported mean range */
+#define TIOVXDLPREPROC_MEAN_NUMBER 1
+static const gdouble
+    tiovxdlpreproc_mean[TIOVXDLPREPROC_MEAN_NUMBER + 1] = { 0.0, 255.0 };
```
emmadrigal, line 139:
```
static const gdouble tiovxdlpreproc_mean[] = { 0.0, 255.0 };```

File: tests/check/gsttiovxdlpreproc.c [hunk 8]
```diff
@@ -0,0 +1,614 @@
...
+
+/* Supported scale range */
+#define TIOVXDLPREPROC_SCALE_NUMBER 1
+static const gdouble
+    tiovxdlpreproc_scale[TIOVXDLPREPROC_SCALE_NUMBER + 1] = { 0.0, 1.0 };
```
emmadrigal, line 144:
```
static const gdouble tiovxdlpreproc_scale[] = { 0.0, 1.0 };
```

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+
+GST_START_TEST (test_state_change_foreach_upstream_format)
+{
+  TIOVXDLPreProcModeled element = { 0 };
+  uint i = 0;
+
+  gst_tiovx_dl_pre_proc_modeling_init (&element);
+
+  for (i = 0; i < TIOVXDLPREPROC_FORMATS_NUMBER; i++) {
+    g_autoptr (GString) pipeline = g_string_new ("");
```
emmadrigal, line 211: I didn't know about `g_autoptr`, its great you are using it!

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+GST_END_TEST;
+
+GST_START_TEST (test_state_change_dimensions)
+{
+  TIOVXDLPreProcModeled element = { 0 };
+  g_autoptr (GString) pipeline = g_string_new ("");
+  g_autoptr (GString) upstream_caps = g_string_new ("");
+  g_autoptr (GString) downstream_caps = g_string_new ("");
+  GRand *g_rand = NULL;
+  const uint tiovx_preproc_image_max_size_root = 3500;
```
emmadrigal, line 261: Where does this variable come from?

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  g_string_printf (downstream_caps,
+      "application/x-tensor-tiovx,width=%d,height=%d", width, height);
+
+  g_string_printf (pipeline,
+      "videotestsrc ! %s ! tiovxdlpreproc ! %s ! fakesink ",
+      upstream_caps->str, downstream_caps->str);
+
+  test_states_change (pipeline->str);
+
+  g_rand_free (g_rand);
```
emmadrigal, line 304: This should be freed after you stop using it

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  g_string_printf (pipeline,
+      "videotestsrc ! %s ! tiovxdlpreproc ! %s ! fakesink ",
+      upstream_caps->str, downstream_caps->str);
+
+  test_states_change_fail (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_state_change_for_framerate)
```
emmadrigal, line 379: Is framerate something we consider in the dlpreproc?

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+    tiovxdlpreproc_mean[TIOVXDLPREPROC_MEAN_NUMBER + 1] = { 0.0, 255.0 };
+
+/* Supported scale range */
+#define TIOVXDLPREPROC_SCALE_NUMBER 1
+static const gdouble
+    tiovxdlpreproc_scale[TIOVXDLPREPROC_SCALE_NUMBER + 1] = { 0.0, 1.0 };
+
+typedef struct
+{
+  const uint *width;
```
emmadrigal, line 148: Use `gtypes` in this whole file

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+
+GST_START_TEST (test_state_change_for_mean_and_scale_fail)
+{
+  g_autoptr (GString) pipeline = g_string_new ("");
+  g_autoptr (GString) properties = g_string_new ("");
+
+  /* Properties */
+  g_string_printf (properties,
+      "mean-0=%p mean-1=%p mean-2=%p scale-0=%p scale-1=%p scale-2=%p", NULL,
+      NULL, NULL, NULL, NULL, NULL);
```
emmadrigal, line 573: You are testing all properties simultaneously, a test per each might be more appropiate

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,614 @@
...
+  tcase_add_test (tc, test_state_transitions);
+  tcase_add_test (tc, test_state_transitions_fail);
+  tcase_add_test (tc, test_state_change_foreach_upstream_format);
+  tcase_add_test (tc, test_state_change_foreach_upstream_format_fail);
+  tcase_add_test (tc, test_state_change_dimensions);
+  tcase_add_test (tc, test_state_change_dimensions_with_upscale_fail);
+  tcase_add_test (tc, test_state_change_dimensions_with_downscale_fail);
+  tcase_add_test (tc, test_state_change_for_framerate);
+
+  suite_add_tcase (suite, tc_properties);
```
emmadrigal, line 601: Why do we have 2 separate test cases?

tests/check/test_utils.h
```diff
@@ -129,6 +129,25 @@ test_states_change_base (const gchar * pipe_desc, GstStateChangeReturn *state_ch
     gst_object_unref (pipeline);
 }
 
+static void
+test_states_change_base_fail (const gchar * pipe_desc, GstStateChangeReturn *state_change);
+
+static void
+test_states_change_base_fail (const gchar * pipe_desc, GstStateChangeReturn *state_change)
```
emmadrigal, line 136: Why have these separate?

emmadrigal (changes requested):

tests/check/gsttiovxdlpreproc.c
dchvs: I'll do it that way for the non arrays of strings; for those cases I'm gonna need it.

tests/check/gsttiovxdlpreproc.c
dchvs: Yeah, at least it is an upstream input parameter.

tests/check/gsttiovxdlpreproc.c
dchvs: Why not, the test suite can have multiples tests cases? In this case, one it's for the properties and the other it's for generical negotiation stuff with the upstream and downstream elements.

tests/check/meson.build
```diff
@@ -8,6 +8,7 @@ gst_tests = [
   ['libs/gsttiovxtensorbufferpool', false, [gst_internal_dep],  [] ],
   ['gsttiovxmultiscaler', false, [gst_internal_dep],  [] ],
   ['gsttiovxcolorconvert', false, [gst_internal_dep],  [] ],
+  ['gsttiovxdlpreproc', false, [gst_internal_dep],  [] ],
```
emmadrigal, line 11: Keep alphabetical order

tests/check/gsttiovxdlpreproc.c
dchvs: Potentially. Because we could eventually parse only these test cases about a certain topic (about properties in this case) and let us apply a regression of only this topic and so on.

tests/check/gsttiovxdlpreproc.c
```diff
@@ -0,0 +1,750 @@
...
+
+GST_END_TEST;
+
+GST_START_TEST (test_state_change_foreach_data_type_fail)
+{
+  g_autoptr (GString) pipeline = g_string_new ("");
+  g_autoptr (GString) properties = g_string_new ("");
+
+  /* Properties */
+  g_string_printf (properties, "data-type=%p", NULL);
```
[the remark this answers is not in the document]
dchvs: I'll skip these tests as it seems to be testing GObject instead of the plugin and its TIOVX backend.

tests/check/gsttiovxdlpreproc.c
emmadrigal: If we need that, we'll separate them then. for now, keep a single test case

emmadrigal (approved):

PR #116, opened 2021-09-29
Feature/add tiovxisp basic structure

jafet-chaves:


ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+#endif
+
+#include "gsttiovxisp.h"
+
+#include "gst-libs/gst/tiovx/gsttiovx.h"
+#include "gst-libs/gst/tiovx/gsttiovxpad.h"
+#include "gst-libs/gst/tiovx/gsttiovxsimo.h"
+#include "gst-libs/gst/tiovx/gsttiovxutils.h"
+
+#include "tiovx_viss_module.h"
```
michaelgruner, line 73: Should this be a system path instead?

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+#define TIOVX_ISP_SUPPORTED_WIDTH "[1 , 8192]"
+#define TIOVX_ISP_SUPPORTED_HEIGHT "[1 , 8192]"
+
+/* Src caps */
+#define TIOVX_ISP_STATIC_CAPS_SRC \
+  "video/x-raw, "							\
+  "format = (string) " TIOVX_ISP_SUPPORTED_FORMATS_SRC ", "					\
+  "width = " TIOVX_ISP_SUPPORTED_WIDTH ", "					\
+  "height = " TIOVX_ISP_SUPPORTED_HEIGHT ", "					\
+  "framerate = " GST_VIDEO_FPS_RANGE
```
michaelgruner, line 94: align these backslashes

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+
+/* Sink caps */
+#define TIOVX_ISP_STATIC_CAPS_SINK \
+  "video/x-bayer, "							\
+  "format = (string) " TIOVX_ISP_SUPPORTED_FORMATS_SINK ", "					\
+  "width = " TIOVX_ISP_SUPPORTED_WIDTH ", "					\
+  "height = " TIOVX_ISP_SUPPORTED_HEIGHT ", "					\
+  "framerate = " GST_VIDEO_FPS_RANGE
```
michaelgruner, line 102: align these backslashes

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+G_DEFINE_TYPE_WITH_CODE (GstTIOVXISP, gst_tiovx_isp,
+    GST_TYPE_TIOVX_SIMO, GST_DEBUG_CATEGORY_INIT (gst_tiovx_isp_debug,
+        "tiovxisp", 0, "debug category for the tiovxisp element"));
+
+static void
+gst_tiovx_isp_set_property (GObject * object, guint prop_id,
+    const GValue * value, GParamSpec * pspec);
+static void
+gst_tiovx_isp_get_property (GObject * object, guint prop_id,
+    GValue * value, GParamSpec * pspec);
```
michaelgruner, line 139: This is not the standard for declaration. Did you use the indenter?

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+      GST_DEBUG_FUNCPTR (gst_tiovx_isp_compare_caps);
+}
+
+/* Initialize the new element
+ * Initialize instance structure
+ */
+static void
+gst_tiovx_isp_init (GstTIOVXISP * self)
+{
+  self->dcc_config_file = NULL;
```
michaelgruner, line 242: All instance variables should be initialized to a known state.

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+static gboolean
+gst_tiovx_isp_set_dcc_file (GstTIOVXISP * self, const gchar * location)
+{
+  GstState state;
+
+  /* the element must be stopped in order to do this */
+  state = GST_STATE (self);
+  if (state != GST_STATE_READY && state != GST_STATE_NULL) {
+    goto wrong_state;
+  }
```
michaelgruner, line 254: 1. This is not thread safe.
2. Maybe you want to take a look at [GST_PARAM_MUTABLE_READY](https://gstreamer.freedesktop.org/documentation/gstreamer/gstparamspec.html?gi-language=c). But I don't know if this actually performs any protection or it's just a nice way to display stuff on the inspect. See the `sink` property of the `glimagesink` for an example.

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,375 @@
...
+
+  g_free (self->dcc_config_file);
+
+  /* clear the filename if we get a NULL */
+  if (location == NULL) {
+    self->dcc_config_file = NULL;
+  } else {
+    self->dcc_config_file = g_strdup (location);
+  }
```
michaelgruner, line 263: this can be reduced to simply `self->dcc_config_file = g_strdup (location);`. This will handle NULL gracefully.

michaelgruner (changes requested):

ext/tiovx/gsttiovxisp.c
jafet-chaves: I agree, it should be a system path but we are handling in this form everywhere else.

ext/tiovx/gsttiovxisp.c
```diff
@@ -0,0 +1,387 @@
...
+
+  G_OBJECT_CLASS (gst_tiovx_isp_parent_class)->finalize (obj);
+}
+
+static gboolean
+gst_tiovx_isp_set_dcc_file (GstTIOVXISP * self, const gchar * location)
+{
+  GstState state;
+
+  g_return_val_if_fail (self, FALSE);
```
emmadrigal, line 265: location pointer guard

emmadrigal (approved):

rrcarlosrodriguez (approved):

ext/tiovx/gsttiovxisp.c
jafet-chaves: 1. The gst_tiovx_isp_set_property calls GST_OBJECT_LOCK/GST_OBJECT_UNLOCK so the call to the function is already thread safe.
2. Using  GST_PARAM_MUTABLE_READY seems to add this protection already in the property, so there is no need to do the check in the function, see: https://github.com/GStreamer/gst-plugins-base/blob/master/gst/videorate/gstvideorate.c#L276

PR #119, opened 2021-10-05
Feature/add mosaic tests

dchvs:


tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,515 @@
...
+
+/* Supported start time selection */
+#define TIOVXMOSAIC_START_TIME_SELECTION_ARRAY_SIZE 3
+static const gchar
+    * tiovxmosaic_start_time_selection
+    [TIOVXMOSAIC_START_TIME_SELECTION_ARRAY_SIZE] = {
+  "zero",
+  "first",
+  "set",
+};
```
emmadrigal, line 106: Latency and start time are aggregator properties and not mosaic properties, these don't belong in this test

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,515 @@
...
+  const gchar **formats;
+} PadTemplateSink;
+
+typedef struct
+{
+  const guint *width;
+  const guint *height;
+  const guint *framerate;
+  const gchar **formats;
+} PadTemplateSrc;
```
emmadrigal, line 138: Why do we need these replicated structures? One should be enough

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,515 @@
...
+  for (i = 0; i < TIOVXMOSAIC_FORMATS_ARRAY_SIZE; i++) {
+    g_autoptr (GString) pipeline = g_string_new ("");
+    g_autoptr (GString) upstream_caps = g_string_new ("");
+
+    /* Upstream caps */
+    g_string_printf (upstream_caps, "video/x-raw,format=%s",
+        element.sink_pad.formats[i]);
+
+    g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! fakesink ",
+        upstream_caps->str);
```
emmadrigal, line 200: Why not do this in a single step?

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,515 @@
...
+
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! fakesink ",
+      upstream_caps->str);
+
+  test_states_change (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_resolutions_with_upscale_fail)
```
emmadrigal, line 254: This test should be succesfull

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,515 @@
...
+
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! %s ! fakesink ",
+      upstream_caps->str, downstream_caps->str);
+
+  test_states_change_fail (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_resolutions_with_downscale_fail)
```
emmadrigal, line 287: This case can be succesful if the input pads are downscaling

emmadrigal (changes requested):

emmadrigal:
You are missing tests for the properties in the input pads: width, height: startx, starty

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! fakesink ",
+      upstream_caps->str);
+
+  test_states_change (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_resolutions_with_upscale)
```
emmadrigal, line 210: I don't like the name of this test, we aren't upscaling anything. 
`with_larger_output` is more accurate

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! %s ! fakesink ",
+      upstream_caps->str, downstream_caps->str);
+
+  test_states_change (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_resolutions_with_downscale_fail)
+{
```
emmadrigal, line 244: This tests name is innacurate, this should be ` with_smaller_output`

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! %s ! fakesink ",
+      upstream_caps->str, downstream_caps->str);
+
+  test_states_change_fail (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_resolutions_smaller_input_into_background)
```
emmadrigal, line 276: `test_downscaled_input` is more accurate

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+
+  /* Downstream caps */
+  g_string_printf (downstream_caps, "video/x-raw,width=%d,height=%d", width,
+      height);
+
+  g_string_printf (pipeline,
+      "videotestsrc ! %s ! tiovxmosaic %s ! %s ! fakesink ", upstream_caps->str,
+      properties->str, downstream_caps->str);
+
+  test_states_change (pipeline->str);
```
emmadrigal, line 349: This test should fail, the input cannot be upscaled

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+}
+
```
emmadrigal, line 351: Missing test for lower limit of downscaling, windows cannot be smaller than 1/4 than the inputs where they come from

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+  g_string_printf (pipeline, "videotestsrc ! %s ! tiovxmosaic ! fakesink ",
+      upstream_caps->str);
+
+  test_states_change (pipeline->str);
+}
+
+GST_END_TEST;
+
+GST_START_TEST (test_request_random_number_of_pads)
```
emmadrigal, line 422: Wrong test name, this doesn't test anything randomly, it tests all cases

tests/check/gsttiovxmosaic.c
```diff
@@ -0,0 +1,497 @@
...
+  tcase_add_test (tc, test_resolutions);
+  tcase_add_test (tc, test_resolutions_with_upscale);
+  tcase_add_test (tc, test_resolutions_with_downscale_fail);
+  tcase_add_test (tc, test_resolutions_smaller_input_into_background);
+  tcase_add_test (tc, test_resolutions_larger_input_into_background);
+  tcase_add_test (tc, test_resolutions_random_startx_starty);
+  tcase_add_test (tc, test_framerate);
+  tcase_add_test (tc, test_request_random_number_of_pads);
+  tcase_add_test (tc, test_property_target);
+
```
emmadrigal, line 493: Missing tests for:
1. Input and output pad have a different format (should fail)
2. Input pads have different formats between them (should fail)
3. Background 
4. Background pad has a different size than output (should fail)
5. Background pad has different format than the input, background pad has different format than the output (both should fail)

emmadrigal (changes requested):

tests/check/gsttiovxmosaic.c
dchvs: This test pases because the element reconfigure to not upscale

PR #121, opened 2021-10-07
Feature/implement tiovxldc element

m-herrera:


ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+  if ((video_info1.width == video_info2.width) &&
+      (video_info1.height == video_info2.height) &&
+      (video_info1.finfo->format == video_info2.finfo->format)
+      ) {
+    ret = TRUE;
+  }
+
+out:
+  return ret;
+}
```
jafet-chaves, line 535: This function will always return TRUE

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+
+#include "gst-libs/gst/tiovx/gsttiovx.h"
+#include "gst-libs/gst/tiovx/gsttiovxpad.h"
+#include "gst-libs/gst/tiovx/gsttiovxsimo.h"
+#include "gst-libs/gst/tiovx/gsttiovxutils.h"
+
+#include "tiovx_ldc_module.h"
+
+#define DEFAULT_NUM_CHANNELS 1
+#define SONY_IMX390_UB953_D3 0
```
rrcarlosrodriguez, line 76: This is never used, remove it

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+/* Properties definition */
+enum
+{
+  PROP_0,
+  PROP_DCC_CONFIG_FILE,
+  PROP_SENSOR_NAME,
+};
+
+/* Formats definition */
+#define TIOVX_LDC_SUPPORTED_FORMATS_SRC "{ GRAY8, GRAY16_LE, NV12, UYVY }"
```
rrcarlosrodriguez, line 90: Did you test all these formats?

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+#define TIOVX_LDC_SUPPORTED_FORMATS_SINK "{ GRAY8, GRAY16_LE, NV12, UYVY }"
+#define TIOVX_LDC_SUPPORTED_WIDTH "[1 , 8192]"
```
rrcarlosrodriguez, line 92: Did you test min and max resolutions?

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+    self->dcc_config_file = g_strdup (location);
+    self->uri = gst_filename_to_uri (location, NULL);
+  }
+
+  return TRUE;
+
+  /* ERROR */
+wrong_state:
+  GST_WARNING_OBJECT (self,
+      "Changing the `dcc-file' property on tiovxldc when a file is open is not supported.");
```
rrcarlosrodriguez, line 334: Weird message, shouldn't we say `Changing the `dcc-file' property on tiovxldc is supported in NULL state only`

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+  sensorObj = &self->sensorObj;
+
+  /* Initialize general parameters */
+  status = tiovx_querry_sensor (sensorObj);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "tiovx query sensor error: %d", status);
+    goto out;
+  }
+
+  status = tiovx_init_sensor (sensorObj, self->sensor_name);
```
rrcarlosrodriguez, line 371: Deinit is never called

rrcarlosrodriguez: tiovx_deinit_sensor must be called in the deinit_module function, but also in this function in case an error happens after this. Single input-Single output pattern

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+
+  /* Initialize modules */
+  GST_INFO_OBJECT (self, "Initializing ldc object");
+  status = tiovx_ldc_module_init (context, ldc, sensorObj);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Module init failed with error: %d", status);
+    goto out;
+  }
+  ret = TRUE;
+out:
```
rrcarlosrodriguez, line 424: Handle better errors here, you have to deinit the sensor is module_init fails

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+  const gchar *target = NULL;
+
+  g_return_val_if_fail (simo, FALSE);
+  g_return_val_if_fail (context, FALSE);
+  g_return_val_if_fail (graph, FALSE);
+
+  self = GST_TIOVX_LDC (simo);
+
+  GST_OBJECT_LOCK (GST_OBJECT (self));
+  // target = target_id_to_target_name (self->target_id);
```
rrcarlosrodriguez, line 479: Implement target property

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+    goto out;
+  }
+
+  status = tiovx_ldc_module_deinit (ldc);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "Module deinit failed with error: %d", status);
+    ret = FALSE;
+    goto out;
+  }
+out:
```
rrcarlosrodriguez, line 565: Deinit sensor here

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+  }
+
+  for (l = src_caps_list; l != NULL; l = l->next) {
+    GstCaps *src_caps = (GstCaps *) l->data;
+    GstStructure *src_st = gst_caps_get_structure (src_caps, 0);
+    GstCaps *new_caps = gst_caps_fixate (gst_caps_ref (src_caps));
+    GstStructure *new_st = gst_caps_get_structure (new_caps, 0);
+    const GValue *vwidth = NULL, *vheight = NULL, *vformat = NULL;
+
+    vwidth = gst_structure_get_value (src_st, "width");
```
rrcarlosrodriguez, line 612: Put a TODO message here just to make sure we know where to add logic to enable different input/output resolutions (waiting for TI's input here)

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,646 @@
...
+  "framerate = " GST_VIDEO_FPS_RANGE
+
+/* Pads definitions */
+static GstStaticPadTemplate sink_template = GST_STATIC_PAD_TEMPLATE ("sink",
+    GST_PAD_SINK,
+    GST_PAD_ALWAYS,
+    GST_STATIC_CAPS (TIOVX_LDC_STATIC_CAPS_SINK)
+    );
+
+static GstStaticPadTemplate src_template = GST_STATIC_PAD_TEMPLATE ("src_%u",
```
rrcarlosrodriguez, line 118: We have to limit the number of pads to create, We are not doing it yet

jafet-chaves (changes requested):

ext/tiovx/gsttiovxldc.c
m-herrera: I did, the original design included  YUYV but it is not supported by the tiovx node so I removed that one

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,801 @@
...
+
+static void
+gst_tiovx_ldc_finalize (GObject * obj)
+{
+
+  GstTIOVXLDC *self = GST_TIOVX_LDC (obj);
+
+  GST_LOG_OBJECT (self, "finalize");
+
+  g_free (self->dcc_config_file);
```
jafet-chaves, line 796: Set to NULL after free

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+
+#include "tiovx_ldc_module.h"
+
+#define DEFAULT_NUM_CHANNELS 1
+#define GST_TYPE_TIOVX_LDC_TARGET (gst_tiovx_ldc_target_get_type())
+#define DEFAULT_TIOVX_LDC_TARGET TIVX_TARGET_VPAC_LDC_ID
+#define MAX_SRC_PADS 1
+
+static const int input_param_id = 6;
+static const int output_param_id_start = 7;
```
michaelgruner, line 81: Please follow the standard chosen by the team. I can't recall if we were going to use macros or constants, but let's be consistent and avoid mixing them.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+
+static void gst_tiovx_ldc_finalize (GObject * obj);
+
+static const gchar *target_id_to_target_name (gint target_id);
+
+static GstPad *gst_tiovx_ldc_request_new_pad (GstElement * element,
+    GstPadTemplate * templ, const gchar * name, const GstCaps * caps);
+
+static gboolean
+gst_tiovx_ldc_set_dcc_file (GstTIOVXLDC * src, const gchar * location);
```
michaelgruner, line 211: Follow the standard of the rest of the declarations.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  GstPadTemplate *src_temp = NULL;
+  GstPadTemplate *sink_temp = NULL;
+
+  gobject_class = G_OBJECT_CLASS (klass);
+  gstelement_class = GST_ELEMENT_CLASS (klass);
+  gsttiovxsimo_class = GST_TIOVX_SIMO_CLASS (klass);
+
+  gst_element_class_set_details_simple (gstelement_class,
+      "TIOVX LDC",
+      "Filter",
```
michaelgruner, line 229: Always check for existing similar elements in order to choose this. For example, if I inspect the `gleffects_fisheye` it suggests `Filter/Effect/Video` .

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  g_object_class_install_property (gobject_class, PROP_TARGET,
+      g_param_spec_enum ("target", "Target",
+          "TIOVX target to use by this element",
+          GST_TYPE_TIOVX_LDC_TARGET,
+          DEFAULT_TIOVX_LDC_TARGET,
+          G_PARAM_READWRITE | GST_PARAM_CONTROLLABLE | G_PARAM_STATIC_STRINGS));
+
+  gsttiovxsimo_class->init_module =
+      GST_DEBUG_FUNCPTR (gst_tiovx_ldc_init_module);
+
```
michaelgruner, line 269: I suggest collapsing these spaces, these are related and is common to see them together.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+/* Initialize the new element
+ * Initialize instance structure
+ */
+static void
+gst_tiovx_ldc_init (GstTIOVXLDC * self)
+{
+  self->dcc_config_file = NULL;
+  self->sensor_name = NULL;
+  self->target_id = 0;
+  self->num_src_pads = 0;
```
michaelgruner, line 310: Initialize the object as well.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+gst_tiovx_ldc_set_dcc_file (GstTIOVXLDC * self, const gchar * location)
+{
+  g_free (self->dcc_config_file);
+
+  /* clear the filename if we get a NULL */
+  if (location == NULL) {
+    self->dcc_config_file = NULL;
+  } else {
+    self->dcc_config_file = g_strdup (location);
+  }
```
michaelgruner, line 376: g_strdup supports NULL, so you can leave these part as simply `self->dcc_config_file = g_strudp (location)`.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+      break;
+    default:
+      G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
+      break;
+  }
+  GST_OBJECT_UNLOCK (self);
+}
+
+static gboolean
+gst_tiovx_ldc_set_dcc_file (GstTIOVXLDC * self, const gchar * location)
```
michaelgruner, line 367: Always use pointer guards of methods we create (and call) ourselves.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  GST_LOG_OBJECT (self, "set_property");
+
+  GST_OBJECT_LOCK (self);
+  switch (prop_id) {
+    case PROP_DCC_CONFIG_FILE:
+      gst_tiovx_ldc_set_dcc_file (self, g_value_get_string (value));
+      break;
+    case PROP_SENSOR_NAME:
+      g_free (self->sensor_name);
+      self->sensor_name = g_value_dup_string (value);
```
michaelgruner, line 328: not how this is exactly what `gst_tiovx_ldc_set_dcc_file` does. It smells like refactoring.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  sensorObj = &self->sensorObj;
+
+  /* Initialize general parameters */
+  status = tiovx_querry_sensor (sensorObj);
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "tiovx query sensor error: %d", status);
+    goto out;
+  }
+
+  status = tiovx_init_sensor (sensorObj, self->sensor_name);
```
michaelgruner, line 414: This is not thread safe. protect `self->sensor_name`.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  if (VX_SUCCESS != status) {
+    GST_ERROR_OBJECT (self, "tiovx init sensor error: %d", status);
+    goto deinit_sensor;
```
michaelgruner, line 417: I don't think deinit should be called if init failed.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,852 @@
...
+  for (i = 0; i < gst_caps_get_size (src_caps); i++) {
+    GstStructure *src_st = gst_caps_get_structure (src_caps, i);
+    GstStructure *template_st = gst_caps_get_structure (template_caps, 0);
+    const GValue *vwidth = NULL, *vheight = NULL;
+
+    vwidth = gst_structure_get_value (template_st, "width");
+    vheight = gst_structure_get_value (template_st, "height");
+
+    gst_structure_set_value (src_st, "width", vwidth);
+    gst_structure_set_value (src_st, "height", vheight);
```
michaelgruner, line 750: Input and output resolutions are independent. We should return the full range instead.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,821 @@
...
+  "height = " TIOVX_LDC_SUPPORTED_HEIGHT ", "			\
+  "framerate = " GST_VIDEO_FPS_RANGE
+
+/* Sink caps */
+#define TIOVX_LDC_STATIC_CAPS_SINK 				\
+  "video/x-raw, "						\
+  "format = (string) " TIOVX_LDC_SUPPORTED_FORMATS_SINK ", "	\
+  "width = " TIOVX_LDC_SUPPORTED_WIDTH ", "			\
+  "height = " TIOVX_LDC_SUPPORTED_HEIGHT ", "			\
+  "framerate = " GST_VIDEO_FPS_RANGE
```
michaelgruner, line 133: These are exactly the same. Please use a single one.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,821 @@
...
+  if (filter) {
+    sink_caps = gst_caps_intersect (template_caps, filter);
+  } else {
+    sink_caps = gst_caps_copy (template_caps);
+  }
+  gst_caps_unref (template_caps);
+
+  for (l = src_caps_list; l != NULL; l = l->next) {
+    GstCaps *src_caps = gst_caps_copy ((GstCaps *) l->data);
+    GstStructure *src_st = gst_caps_get_structure (src_caps, 0);
```
michaelgruner, line 768: How do you know this only has one structure? It may have several.

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,821 @@
...
+
+    if (gst_caps_can_intersect (sink_caps, src_caps)) {
+      tmp = gst_caps_intersect (sink_caps, src_caps);
+      gst_caps_unref (sink_caps);
+      gst_caps_unref (src_caps);
+      sink_caps = tmp;
+    } else {
+      gst_caps_unref (src_caps);
+      break;
+    }
```
michaelgruner, line 782: Why do you handle a special case when there is no intersection? if there is no intersection in one of the pads, we must fail (aka return the failed intersection).

ext/tiovx/gsttiovxldc.c
```diff
@@ -0,0 +1,821 @@
...
+  GST_DEBUG_OBJECT (simo,
+      "Computing src caps based on sink caps %" GST_PTR_FORMAT " and filter %"
+      GST_PTR_FORMAT, sink_caps, filter);
+
+  template_caps = gst_static_pad_template_get_caps (&src_template);
+  template_st = gst_caps_get_structure (template_caps, 0);
+  vwidth = gst_structure_get_value (template_st, "width");
+  vheight = gst_structure_get_value (template_st, "height");
+
+  src_caps = gst_caps_intersect (template_caps, sink_caps);
```
michaelgruner, line 720: I don't like doing this intersection when the "width" and "height" fields haven't been removed. Right now is not problematic because we know both sink and src template overlap, but it may not be the case. I prefer taking a similar approach as you did with `get_sink_caps`.

michaelgruner (changes requested):

ext/tiovx/gsttiovxisp.c
```diff
@@ -151,16 +151,14 @@ gst_tiovx_isp_target_get_type (void)
   "video/x-raw, "                                            \
   "format = (string) " TIOVX_ISP_SUPPORTED_FORMATS_SRC ", "  \
   "width = " TIOVX_ISP_SUPPORTED_WIDTH ", "                  \
-  "height = " TIOVX_ISP_SUPPORTED_HEIGHT ", "                \
-  "framerate = " GST_VIDEO_FPS_RANGE
+  "height = " TIOVX_ISP_SUPPORTED_HEIGHT
```
michaelgruner, line 154: Why did we remove the framerate. This sounds like a hack.

michaelgruner (approved):

PR #129, opened 2021-10-18
Feature/refactor graph configuration method

dchvs99:


gst-libs/gst/tiovx/gsttiovxutils.c
```diff
@@ -462,3 +462,40 @@ gst_format_to_tivx_raw_format (const gchar * gst_format)
...
+  g_return_val_if_fail (refs_list, VX_FAILURE);
+  g_return_val_if_fail (VX_SUCCESS ==
+      vxGetStatus ((vx_reference) graph), VX_FAILURE);
+  g_return_val_if_fail (VX_SUCCESS ==
+      vxGetStatus ((vx_reference) node), VX_FAILURE);
+
+  parameter = vxGetParameterByIndex (node, node_parameter_index);
+  status = vxAddParameterToGraph (graph, parameter);
+  if (VX_SUCCESS != status) {
+    GST_ERROR ("Add parameter to graph failed  %" G_GINT32_FORMAT, status);
```
emmadrigal, line 487: We can't return an `GST_ERROR` this won't give a user information about the class or be able to filter it by class name.

This method should receive a `GstDebugCategory *` and the calling `GObject` and then you can call:
https://gstreamer.freedesktop.org/documentation/gstreamer/gstinfo.html?gi-language=c#GST_CAT_ERROR_OBJECT

gst-libs/gst/tiovx/gsttiovxutils.h
```diff
@@ -182,4 +185,15 @@ gsize gst_tiovx_get_size_from_exemplar (vx_reference exemplar);
  */
 void gst_tiovx_init_debug (void);
 
+/**
+ * add_graph_parameter_by_node_index:
+ *
```
emmadrigal, line 190: You need to document all variables

emmadrigal (changes requested):

Issue #128, opened 2021-10-18
tiovxdlcolorblend not working when followed by tiovxmosaic


aniket-l:
Following gives a segmentation fault after first frame:
```
gst-launch-1.0 \
tiovxdlcolorblend num-classes=8 data-type=3 name=blend \
videotestsrc pattern=18 num-buffers=100 is-live=true ! "video/x-raw,format=NV12,width=1280,height=720" ! queue  ! blend.sink \
multifilesrc num-buffers=100 loop=true location=baboon_mask_320x240.bin blocksize=76800 ! "application/x-tensor-tiovx, data-type=3, num-dims=3, tensor-width=320, tensor-height=240" ! blend.tensor  \
blend.src ! "video/x-raw,format=NV12,width=1280,height=720" ! queue ! mosaic.sink_0  \
tiovxmosaic name=mosaic sink_0::startx=320   sink_0::starty=180   !  \
"video/x-raw,format=NV12, width=1920, height=1080" ! kmssink sync=false driver-name=tidss
```

PR #131, opened 2021-10-21
Fix/empty exemplar on caps renegotiation

m-herrera:


m-herrera:
This solves #128

gst-libs/gst/tiovx/gsttiovxmiso.c
```diff
@@ -164,7 +164,6 @@ gst_tiovx_miso_pad_finalize (GObject * obj)
   GstTIOVXMisoPadPrivate *priv = gst_tiovx_miso_pad_get_instance_private (self);
 
   if (priv->exemplar) {
-    vxReleaseReference (priv->exemplar);
```
michaelgruner, line 167: Please double check this doesn't leave a leak behind.

michaelgruner (approved):

PR #143, opened 2021-11-10
Feature/support background as property on tiovxmosaic

dchvs:


ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1125,138 @@ target_id_to_target_name (gint target_id)
...
+  if (0 > file_size) {
+    GST_ERROR_OBJECT (self, "File: %s has invalid size", self->background);
+    fclose (background_img_file);
+    goto out;
+  }
+
+  file_buffer = tivxMemAlloc (file_size, TIVX_MEM_EXTERNAL);
+  fread (file_buffer, 1, file_size, background_img_file);
+
+  memcpy ((void *) *addr, (const void *) file_buffer, planes_offset);
```
emmadrigal, line 1212: You can't just copy the memory 1:1.  If the stride doesn't match the width the output will be incorrect.

See the following logic:
https://github.com/TexasInstruments/edgeai-tiovx-modules/blob/main/test/app_common.c#L1415

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1125,138 @@ target_id_to_target_name (gint target_id)
...
+  fseek (background_img_file, 0, SEEK_SET);
+
+  if (0 > file_size) {
+    GST_ERROR_OBJECT (self, "File: %s has invalid size", self->background);
+    fclose (background_img_file);
+    goto out;
+  }
+
+  file_buffer = tivxMemAlloc (file_size, TIVX_MEM_EXTERNAL);
+  fread (file_buffer, 1, file_size, background_img_file);
```
emmadrigal, line 1210: Is there a need for this to be TI memory? the output `addr` is already TI memory

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -348,6 +356,23 @@ static gboolean gst_tiovx_mosaic_deinit_module (GstTIOVXMiso * agg);
...
+    self);
+static gboolean
+gst_tiovx_mosaic_allocate_single_user_data_object (GstTIOVXMosaic * self,
+    GstMemory ** memory, vx_image background_img);
+
+static inline gboolean
+gst_tiovx_mosaic_has_background (const gchar * background)
+{
+  gboolean ret = FALSE;
+
```
emmadrigal, line 369: pointer guard

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -348,6 +356,23 @@ static gboolean gst_tiovx_mosaic_deinit_module (GstTIOVXMiso * agg);
...
     GList * sink_caps_list, GstCaps * src_caps);
 
+static gboolean gst_tiovx_mosaic_allocate_user_data_objects (GstTIOVXMosaic *
+    self);
+static gboolean
+gst_tiovx_mosaic_allocate_single_user_data_object (GstTIOVXMosaic * self,
+    GstMemory ** memory, vx_image background_img);
+
+static inline gboolean
+gst_tiovx_mosaic_has_background (const gchar * background)
```
emmadrigal, line 366: inacurate name

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -76,6 +77,8 @@ static const int input_param_id_start = 3;
 
 static const int window_downscaling_max_ratio = 4;
 
+#define MODULE_MAX_NUM_PLANES 4
```
jafet-chaves, line 80: Keep the standard on new code: static const int

emmadrigal (changes requested):

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -261,12 +263,14 @@ gst_tiovx_mosaic_target_get_type (void)
 }
 
 #define DEFAULT_TIOVX_MOSAIC_TARGET TIVX_TARGET_VPAC_MSC1_AND_2_ID
+#define DEFAULT_TIOVX_MOSAIC_BACKGROUND ""
```
jafet-chaves, line 266: Keep the standard on new code: static const int

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -348,6 +356,23 @@ static gboolean gst_tiovx_mosaic_deinit_module (GstTIOVXMiso * agg);
...
+gst_tiovx_mosaic_allocate_single_user_data_object (GstTIOVXMosaic * self,
+    GstMemory ** memory, vx_image background_img);
+
+static inline gboolean
+gst_tiovx_mosaic_has_background (const gchar * background)
+{
+  gboolean ret = FALSE;
+
+  /* TODO: Validate the background image */
+  ret = (0 != g_strcmp0 ("", background)) ? TRUE : FALSE;
```
jafet-chaves, line 371: Before this it is missing pointer guard for background

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1125,138 @@ target_id_to_target_name (gint target_id)
 
   return value_nick;
 }
+
+static gboolean
+gst_tiovx_mosaic_allocate_single_user_data_object (GstTIOVXMosaic * self,
```
jafet-chaves, line 1130: Missing pointer guards for function that we define and create

jafet-chaves (changes requested):

ext/tiovx/gsttiovxmosaic.c
dchvs: This variable is meant to initialize a pointer of arrays. Therefore it cannot be `static const int` in C, because it will give Variable Length Array (VLA) error by default. See https://www.keil.com/support/man/docs/armcc/armcc_chr1359124950297.htm.

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -126,6 +132,41 @@ static void
 gst_tiovx_mosaic_pad_get_property (GObject * object, guint prop_id,
     GValue * value, GParamSpec * pspec);
 
+vx_uint32 *gst_tiovx_mosaic_learn_planes_offsets (GstTIOVXMosaic * self,
```
jafet-chaves, line 135: I propose the following function name:
`gst_tiovx_mosaic_compute_planes_offsets`

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -126,6 +132,41 @@ static void
...
+    vx_image img);
+/*static*/ vx_uint32 *
```
jafet-chaves, line 137: Remove commented code

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -126,6 +132,41 @@ static void
...
+gst_tiovx_mosaic_learn_planes_offsets (GstTIOVXMosaic * self, vx_image img)
```
jafet-chaves, line 138: I dont understand this, it does not seem be doing anything useful

jafet-chaves (approved):

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -411,7 +434,10 @@ gst_tiovx_mosaic_init (GstTIOVXMosaic * self)
   memset (&self->obj, 0, sizeof (self->obj));
 
   self->target_id = DEFAULT_TIOVX_MOSAIC_TARGET;
+  self->background = g_strdup (default_tiovx_mosaic_background);
   self->has_background = FALSE;
+  self->user_data_allocator = g_object_new (GST_TYPE_TIOVX_ALLOCATOR, NULL);
```
emmadrigal, line 439: You are never calling `g_object_unref` on this object

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+  guint i = 0;
+  gint w = 0;
+  gint h = 0;
+  vx_rectangle_t rectangle = { 0 };
+  vx_imagepatch_addressing_t image_addr = { 0 };
+  vx_map_id map_id = 0;
+  guint planes_offset = 0;
+  gint stride_length = 0;
+
+  g_return_val_if_fail (self, FALSE);
```
emmadrigal, line 1159: memory, background_img pointer guards

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+  vx_uint32 plane_sizes[MODULE_MAX_NUM_PLANES] = { 0 };
+  guint num_planes = 0;
+  vx_size data_size = 0;
+  GstTIOVXMemoryData *ti_memory = NULL;
+  FILE *background_img_file = NULL;
+  gint file_size = 0;
+  void *file_buffer = NULL;
+  guint i = 0;
+  gint w = 0;
+  gint h = 0;
```
emmadrigal, line 1152: Use more representative names, `width`, and `height` are much better

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+    goto out;
+  }
+  GST_DEBUG_OBJECT (self, "Height for background image: %d", h);
+
+
+  /* Alloc GStreamer memory */
+  *memory =
+      gst_allocator_alloc (GST_ALLOCATOR (self->user_data_allocator), data_size,
+      NULL);
+  if (!*memory) {
```
emmadrigal, line 1205: `NULL == *memory` is how we are checking for null pointers

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+  /* Organize the memory per plane pointers */
+  for (i = 0; i < num_planes; i++) {
+    guint j = 0;
+    gint w0 = 0;
+
+    rectangle.start_x = 0;
+    rectangle.start_y = 0;
+    rectangle.end_x = w;
+    rectangle.end_y = h;
+    status =
```
emmadrigal, line 1244: Do you need to make this for very plane?

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+
+    addr[i] = ((char *) ti_memory->mem_ptr.host_ptr + planes_offset);
+
+    if (data_size == file_size) {
+      GST_DEBUG_OBJECT (self,
+          "Got background image with no padding; width matches stride");
+      fread (file_buffer, 1, plane_sizes[i], background_img_file);
+      memcpy ((void *) addr[i], (const void *) file_buffer, plane_sizes[i]);
+
+      file_buffer = (char *) file_buffer + planes_offset;
```
emmadrigal, line 1256: This check should be outside of this loop, you just need to copy image_size from file to `addr[0]` once since memory is contiguous between planes

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+    } else {
+      GST_DEBUG_OBJECT (self,
+          "Got background image with padding; width doesn't match stride");
+      w0 = ((image_addr.dim_x * image_addr.stride_x) / image_addr.step_x);
+      stride_length = image_addr.dim_y / image_addr.step_y;
```
emmadrigal, line 1261: This is not stride lenght. Plane rows is what you are looking for

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+
+      for (j = 0; j < stride_length; j++) {
+        fread (file_buffer, 1, w0, background_img_file);
+        memcpy ((void *) addr[i], (const void *) file_buffer, w0);
+
+        addr[i] = (char *) addr[i] + image_addr.stride_y;
+        file_buffer = (char *) file_buffer + w0;
+      }
+
+      /* Return pointer to plain base */
```
emmadrigal, line 1271: Typo `plane`

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
...
+  if (!ret && *memory) {
+    gst_memory_unref (*memory);
+    *memory = NULL;
+  }
+
+  return ret;
+}
+
+static gboolean
+gst_tiovx_mosaic_allocate_user_data_objects (GstTIOVXMosaic * self)
```
emmadrigal, line 1303: You are not allocating any [user data objects](https://software-dl.ti.com/jacinto7/esd/processor-sdk-rtos-jacinto7/latest/exports/docs/tiovx/docs/user_guide/vx__khr__user__data__object_8h.html#a863903e2890171d914598216e0727d73) in this function. Change the name to something more accurate

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1070,3 +1131,202 @@ target_id_to_target_name (gint target_id)
 
   return value_nick;
 }
+
+static gboolean
+gst_tiovx_mosaic_allocate_background_image (GstTIOVXMosaic * self,
```
emmadrigal, line 1136: This function can have a better name, you are not only allocating the memory, but also loading the data to the memory

emmadrigal (changes requested):

ext/tiovx/gsttiovxmosaic.c
dchvs: It's because of the VX image patch that iterates per plane index. See the `plane_index: vx_uint32` parameter below: 

`VX_API_ENTRY vx_status VX_API_CALL vxMapImagePatch(vx_image image, const vx_rectangle_t *rect, vx_uint32 plane_index, vx_map_id *map_id, vx_imagepatch_addressing_t *addr, void **ptr, vx_enum usage, vx_enum mem_type, vx_uint32 flags);`

ext/tiovx/gsttiovxmosaic.c
dchvs: These are outputs parameters

emmadrigal: `*memory` is an output paramter, `memory` is an input
also background_img is a pointer who also needs a pointer guard

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,201 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+
+  /* Organize the memory per plane pointers */
+  for (i = 0; i < num_planes; i++) {
+    guint j = 0;
+    gint width_per_plane = 0;
+
+    rectangle.start_x = 0;
+    rectangle.start_y = 0;
+    rectangle.end_x = width;
+    rectangle.end_y = height;
```
emmadrigal, line 1261: Rectangle doesn't change between planes, there is no need to set it every iteration

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,201 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+    status =
+        vxMapImagePatch (background_img, &rectangle, i, &map_id, &image_addr,
+        &file_buffer, VX_WRITE_ONLY, VX_MEMORY_TYPE_HOST, VX_NOGAP_X);
```
emmadrigal, line 1264: `file_buffer` is the wrong name for this variable, it comes from the memory rather than disk

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,201 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+    } else {
+      GST_DEBUG_OBJECT (self,
+          "Got background image with padding; width doesn't match stride");
+      width_per_plane =
+          ((image_addr.dim_x * image_addr.stride_x) / image_addr.step_x);
+      plane_rows = image_addr.dim_y / image_addr.step_y;
+
+      for (j = 0; j < plane_rows; j++) {
+        fread (file_buffer, 1, width_per_plane, background_img_file);
+        memcpy ((void *) addr[i], (const void *) file_buffer, width_per_plane);
```
emmadrigal, line 1281: Why are we still performing the copy twice?
`fread (addr[i], 1, plane_sizes[i], background_img_file);` 
whould work withtout having an intermediate file_buffer

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,201 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+    status =
+        vxMapImagePatch (background_img, &rectangle, i, &map_id, &image_addr,
+        &file_buffer, VX_WRITE_ONLY, VX_MEMORY_TYPE_HOST, VX_NOGAP_X);
+
+    addr[i] = ((char *) ti_memory->mem_ptr.host_ptr + planes_offset);
+
+    if (data_size == file_size) {
+      GST_DEBUG_OBJECT (self,
+          "Got background image with no padding; width matches stride");
+      fread (addr[i], 1, plane_sizes[i], background_img_file);
```
emmadrigal, line 1271: This check should be performed outside the plane loop, we should just copy once from:
background_img_file to addr[0] with a size of data_size/file_size

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,201 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+      width_per_plane =
+          ((image_addr.dim_x * image_addr.stride_x) / image_addr.step_x);
+      plane_rows = image_addr.dim_y / image_addr.step_y;
+
+      for (j = 0; j < plane_rows; j++) {
+        fread (file_buffer, 1, width_per_plane, background_img_file);
+        memcpy ((void *) addr[i], (const void *) file_buffer, width_per_plane);
+
+        addr[i] = (char *) addr[i] + image_addr.stride_y;
+        file_buffer = (char *) file_buffer + width_per_plane;
```
emmadrigal, line 1284: This `file_buffer` is being overwritten on every frame

emmadrigal (changes requested):

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -777,6 +832,16 @@ gst_tiovx_mosaic_deinit_module (GstTIOVXMiso * agg)
...
 
+  if (self->background_image_memory) {
+    gst_memory_unref (self->background_image_memory);
+    self->background_image_memory = NULL;
+  }
+
+  if (self->user_data_allocator) {
+    gst_object_unref (self->user_data_allocator);
+    self->user_data_allocator = NULL;
+  }
```
emmadrigal, line 843: Since you created this during the `init` this should be unrefed during the `finalize`.

If after moving it, OpenVX has some unreleased memory due to the sequence you can leave it here

ext/tiovx/gsttiovxmosaic.c
```diff
@@ -1080,3 +1147,202 @@ gst_tiovx_mosaic_get_next_time (GstAggregator * agg)
...
+
+  /* Organize the memory per plane pointers */
+  for (i = 0; i < num_planes; i++) {
+    guint j = 0;
+    gint width_per_plane = 0;
+
+    status =
+        vxMapImagePatch (background_img, &rectangle, i, &map_id, &image_addr,
+        &img_buffer, VX_WRITE_ONLY, VX_MEMORY_TYPE_HOST, VX_NOGAP_X);
```
emmadrigal, line 1266: This can be moved inside the `data_size != file_size` if

emmadrigal (changes requested):

emmadrigal (approved):

PR #170, opened 2021-12-29
mosaic: setting default target to VPAC_MSC2

shyam-j:
- This helps prevent a race-around condition when two processess
  creates the same MSC target instance. It is typical that atleast
  two instances of MSC will be running in separate processes. So
  by default let us not enable both MSC instance in mosaic element.

  The proper fix is however in the OpenVx driver, this will be fixed
  post 8.1

Signed-off-by: Shyam Jagannathan <shyam.jagannathan@ti.com>


shyam-j:
This doesn't seem to fix the issue. Closing the PR.

PR #189, opened 2022-02-15
Feature/isp multichannel support

emmadrigal:


rrcarlosrodriguez (changes requested):
- Check for memory leaks
- Test all ISP properties
- Run destructive tests with caps mismatch, resolutions not supported, non-tiovx buffers, etc

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -1509,6 +1501,33 @@ gst_tiovx_simo_process_graph (GstTIOVXSimo * self)
   return ret;
 }
 
+static gboolean
+gst_tiovx_simo_modules_deinit (GstTIOVXSimo * self)
```
rrcarlosrodriguez, line 1505: Pointer guard for self

gst-libs/gst/tiovx/gsttiovxsimo.c
```diff
@@ -1509,6 +1501,33 @@ gst_tiovx_simo_process_graph (GstTIOVXSimo * self)
...
+  g_list_free_full (priv->queueable_objects, g_object_unref);
+  priv->queueable_objects = NULL;
+
+  if (NULL == klass->deinit_module) {
+    GST_ERROR_OBJECT (self, "Subclass did not implement deinit_module method");
+    goto release_graph;
+  }
+  ret = klass->deinit_module (self);
+  if (!ret) {
+    GST_ERROR_OBJECT (self, "Subclass deinit module failed");
```
rrcarlosrodriguez, line 1520: goto release_graph, otherwise ret will be overwritten

rrcarlosrodriguez (approved):
Approved with comments

PR #199, opened 2022-02-24
Add delay element

emmadrigal:


ext/tiovx/gsttiovxdelay.c
```diff
@@ -0,0 +1,238 @@
...
+{
+  GstElementClass *gstelement_class = GST_ELEMENT_CLASS (klass);
+  GstBaseTransformClass *base_transform_class =
+      GST_BASE_TRANSFORM_CLASS (klass);
+  GObjectClass *gobject_class = (GObjectClass *) klass;
+
+  gst_element_class_set_details_simple (gstelement_class,
+      "TIOVX Delay",
+      "Generic",
+      "Delays buffers by a given amount. The first buffer will be replicated will the delay is reached.",
```
rrcarlosrodriguez, line 125: typo: while

rrcarlosrodriguez (approved):
With a simple comment.

Also, we need to document with special care the deadlock that might be caused if using a delay size >= the buffer pool on the upstream pad. The delay element will hold the buffers and upstream element won't have any buffer available to push downstream and the pipeline will get stuck, Please test and document this accordingly

m-herrera (approved):

PR #203, opened 2022-03-08
Feature/implement pyramid element

m-herrera:


ext/tiovx/gsttiovx.c
```diff
@@ -92,6 +93,13 @@ ti_ovx_init (GstPlugin * plugin)
     goto out;
   }
 
+  ret = gst_element_register (plugin, "tiovxdemux", GST_RANK_NONE,
+      GST_TYPE_TIOVX_DEMUX);
+  if (!ret) {
+    GST_ERROR ("Failed to register the tiovxdemux element");
+    goto out;
+  }
+
```
emmadrigal, line 102: Why did you move this up? this change should be a part of your PR

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,589 @@
...
+  GstBaseTransformClass *gstbasetransform_class = NULL;
+
+  gstelement_class = GST_ELEMENT_CLASS (klass);
+  gsttiovxsiso_class = GST_TIOVX_SISO_CLASS (klass);
+  gstbasetransform_class = GST_BASE_TRANSFORM_CLASS (klass);
+
+  gst_element_class_set_details_simple (gstelement_class,
+      "TIOVX Pyramid",
+      "Filter/Converter/Video",
+      "Converts video frames to a vx_pyramid representation using the TIOVX Modules API",
```
emmadrigal, line 184: Not `vx_pyramid` we are abstracting the OpenVX implementation through the modules and then through our own modules, the user shouldn't need to know that this is a vx_pyramid. Just `pyramid` should be fine

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,589 @@
...
+  if (GST_VIDEO_FORMAT_UNKNOWN == format) {
+    GST_ERROR_OBJECT (self, "Format not found in pyramid caps");
+    goto exit;
+  }
+
+  /* Configure pyramid */
+  pyramid = &self->obj;
+  pyramid->num_channels = num_channels;
+  pyramid->input.bufq_depth = num_channels;
+  pyramid->output.bufq_depth = num_channels;
```
emmadrigal, line 294: Move these to the `Configure input` and `Configure output` sections

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,589 @@
...
+    }
+    if (current_caps) {
+      GstCaps *tmp = current_caps;
+      current_caps = gst_caps_intersect (result_caps, current_caps);
+      gst_caps_unref (tmp);
+      structure = gst_caps_get_structure (current_caps, 0);
+      vscale = gst_structure_get_value (structure, "scale");
+      gst_tiovx_pyramid_set_max_levels (self, vwidth, vheight, vscale,
+          &vlevels);
+    }
```
emmadrigal, line 358: Shouldn't we fail on the `else` case?

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,589 @@
...
+
+static void
+gst_tiovx_pyramid_set_max_levels (GstTIOVXPyramid * self, const GValue * vwidth,
+    const GValue * vheight, const GValue * vscale, GValue * vlevels)
+{
+  guint i = 0;
+  gdouble max_scale = 0;
+  gint max_width = 0, max_height = 0;
+  gint max_levels = MODULE_MAX_NUM_PYRAMIDS, min_levels = 1;
+
```
emmadrigal, line 395: pointer guards

emmadrigal (changes requested):

ext/tiovx/gsttiovx.c
m-herrera: I moved it to keep the alphabetical order

ext/tiovx/gsttiovxpyramid.c
m-herrera: I added a debug message indicating that there were no current caps on the src pad, but I don't think it is necessary to fail since it can still transform the caps based on the template

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,706 @@
...
+/* Properties definition */
+enum
+{
+  PROP_0,
+  PROP_TARGET,
+};
+
+/* Formats definition */
+#define TIOVX_PYRAMID_SUPPORTED_FORMATS "{GRAY8, GRAY16_LE}"
+#define TIOVX_PYRAMID_SUPPORTED_WIDTH "[1 , 1920]"
```
rrcarlosrodriguez, line 115: Same here

rrcarlosrodriguez (changes requested):

ext/tiovx/gsttiovxpyramid.c
m-herrera: I removed the validation altogether, since gst_pad_peer_query_caps is never NULL as seen [here](https://gstreamer.freedesktop.org/documentation/gstreamer/gstpad.html?gi-language=c#gst_pad_peer_query_caps)

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,701 @@
...
+    if (!current_caps) {
+      current_caps = gst_pad_peer_query_caps (base->srcpad, NULL);
+    }
+    tmp = current_caps;
+    current_caps = gst_caps_intersect (result_caps, current_caps);
+    gst_caps_unref (tmp);
+    structure = gst_caps_get_structure (current_caps, 0);
+    vscale = gst_structure_get_value (structure, "scale");
+    gst_tiovx_pyramid_set_max_levels (self, vwidth, vheight, vscale, &vlevels);
+    gst_caps_unref (current_caps);
```
emmadrigal, line 434: `current_caps = NULL`

ext/tiovx/gsttiovxpyramid.c
```diff
@@ -0,0 +1,701 @@
...
+    gst_structure_set_value (result_structure, "format", vformat);
+    if (G_IS_VALUE (&vlevels) && GST_PAD_SINK == direction) {
+      gst_structure_set_value (result_structure, "levels", &vlevels);
+    }
+  }
+
+  if (filter) {
+    tmp = result_caps;
+    result_caps = gst_caps_intersect (result_caps, filter);
+    gst_caps_unref (tmp);
```
emmadrigal, line 458: `tmp = NULL`

emmadrigal (approved):

PR #218, opened 2022-05-25
Avoid input reference on same output return

m-herrera:


m-herrera:
The input buffer is not unrefed by BaseTransform if it is equal to the output buffer

emmadrigal (approved):

PR #375, opened 2025-11-26
Tiovxmultiscaler: Add test cases for simultaneous processing and cropping

PrathamTI:
This commit introduces test cases to verify the
behavior of the tiovxmultiscaler element under various
configurations of simultaneous processing and cropping.

The following scenarios are covered: 
1.Simultaneous processing and cropping are off
2.Simultaneous processing is off and cropping is on
3.Simultaneous processing is on and cropping is off
4.Simultaneous processing and cropping are on

Extensive testing was performed for each of these conditions
to ensure correct functionality.
Test logs are available at: [0]

[0] https://gist.github.com/PrathamTI/72b70f42464b181fa4a4f9dcbbda13b1
Signed-off-by: Pratham Deshmukh p-deshmukh@ti.com

### DIFF
--- a/tests/check/gsttiovxmultiscaler.c
+++ b/tests/check/gsttiovxmultiscaler.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2021] Texas Instruments Incorporated
+ * Copyright (c) [2021-2025] Texas Instruments Incorporated
  *
  * All rights reserved not granted herein.
  *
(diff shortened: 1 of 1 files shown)


tests/check/gsttiovxmultiscaler.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2021] Texas Instruments Incorporated
+ * Copyright (c) [2025] Texas Instruments Incorporated
```
cshilwant:
```suggestion
 * Copyright (c) [2021-2025] Texas Instruments Incorporated
```

cshilwant (changes requested):
good that you handled the year update but it should be in the form of range so as to keep both `start year` - `end year`

cshilwant (approved):

Akshay-Abdar (approved):

PR #380, opened 2026-07-01
timosaic: add UYVY/YUY2 sink pad support

jeevantelukula:
timosaic currently accepts only NV12 on sink pads. This patch expands the sink caps to {NV12, UYVY, YUY2} and handles conversion internally using the existing ARMv8 NEON utilities already used by gstticolorconvert.

UYVY/YUY2 input -> colorConvert_YUVXtoNV12_i8u_o8u_armv8() (pre-allocated buf) -> scaleNNNV12() -> output tile

The intermediate NV12 buffer is allocated once at caps negotiation (not per-frame) following the gsttiscaler.c pre-allocation pattern. The NV12 fast path is unchanged.

Fixes: SITREQ-5709

-------------------

Logs: https://gist.github.com/jeevantelukula/d5d907f16f21b4f12936f89a89fe341b

### DIFF
--- a/ext/ti/gsttimosaic.c
+++ b/ext/ti/gsttimosaic.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) [2023] Texas Instruments Incorporated
+ * Copyright (c) [2023-2026] Texas Instruments Incorporated
  * 
  * All rights reserved not granted herein.
  * 
@@ -65,6 +65,7 @@
 
 #include "gsttimosaic.h"
 #include <edgeai_dl_scaler_armv8_utils.h>
+#include <edgeai_dl_color_convert_armv8_utils.h>
 #include <edgeai_arm_neon_utils.h>
 #include <stdio.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


cshilwant (commented):
@jeevantelukula as discussed, let's try some more benchmarking results for CPU usage comparing `ticolorconvert` vs `glcolorconvert` vs your implementation (which internally still does colorconversion with Arm neon optimization...which ticolorconvert does). want to see how they stack up against each other in terms of actual CPU utilization.

The primary ask in E2E ticket under SITREQ-5709 was lowering down CPU utilization
(for which the ask was to add `UYVY/YUY2` sink pad support)

Also, check with Suren once on what's the expected KPI from customer (in terms of CPU usage)

jeevantelukula:
> @jeevantelukula as discussed, let's try some more benchmarking results for CPU usage comparing `ticolorconvert` vs `glcolorconvert` vs your implementation (which internally still does colorconversion with Arm neon optimization...which ticolorconvert does). want to see how they stack up against each other in terms of actual CPU utilization.
> 
> The primary ask in E2E ticket under SITREQ-5709 was lowering down CPU utilization (for which the ask was to add `UYVY/YUY2` sink pad support)
> 
> Also, check with Suren once on what's the expected KPI from customer (in terms of CPU usage)

@cshilwant  Updated the logic and Benchmarked the pipelines here: https://gist.github.com/jeevantelukula/fccf05aeefc4890da9fa8f5ce6f6c538

cshilwant (approved):

PR #388, opened 2026-09-07
Add tidspkernel and titvm plugins for AM62D speech enhancement and audio classification

PrathamTI:
Adds two new GStreamer elements for the AM62D C7x DSP, enabling audio speech-enhancement (GCRN) and audio-classification (YAMNet, VGGish) pipelines end to end on-device.

- **`tidspkernel`** — offloads STFT, ISTFT, deinterleave, and interleave
    operations to the C7x DSP over RPMsg. A single generic element,
    selected via `msg-type` or auto-detected from the element's name.
    Generalized beyond GCRN-only hardcoding via `selected-model`,
    `model-elems`, and `model-path` (auto-detects `dccrn`/`gtcrn`/`gcrn`/
    `vggish`/`yamnet` by name). `overlap-frames`
    and `chunking-mode` make the overlap-save behavior explicitly
    overridable for models outside the three validated here.

 - **`titvm`** — runs TVM-compiled models (`deploy_lib.so`,
    `deploy_graph.json`, `deploy_param.params`) via the TVM graph executor.
    Prefers the persistent `tvm-model-daemon` over loading the model
    in-process (the C7x compute channel only supports one client at a
    time), with automatic daemon model-switching when the pipeline
    requests a different model than what's currently loaded. Input/output
    tensor shape is auto-detected from `deploy_graph.json`.
    `class-map-path`/`top-k` print live top-k class predictions for
    classification models, off by default and a no-op for GCRN.

  - A shared, refcounted RPMsg channel utility lets multiple
    `tidspkernel` instances in the same pipeline safely share one
    endpoint to the C7x DSP instead of each opening their own.


ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1786 @@
...
+  for (batch_idx = 0; batch_idx < num_batches; batch_idx++) {
+    guint frame_start = batch_idx * kernel->batch_size;
+    guint frames_in_batch = MIN (kernel->batch_size,
+        kernel->window_frames - frame_start);
+    gsize batch_bytes = frames_in_batch * kernel->hop_size * sizeof (gint16);
+    gsize sample_offset = frame_start * kernel->hop_size;
+    struct stft_istft_msg req = { }, resp = { };
+    gsize batch_spectral_bytes;
+
+    dmabuf_sync (kernel->dma_input.dma_buf_fd, DMA_BUF_SYNC_START);
```
v-singh1: The kernel->dma_input and kernel->dma_output buffers are shared state that is accessed without per-element locking. While the shared RPMsg channel is protected by gst_ti_rpmsg_chan_lock(), the DMA buffers themselves are not. In a GStreamer pipeline with multiple threads or concurrent elements sharing the same instance, this can cause: Data corruption, silent failures, DSP kernel receiving wrong input.

// In gstdspkernel.h, add to struct _GstDspKernel:
GMutex dma_mutex;  // Protects dma_input/dma_output access

// In gst_dsp_kernel_init():
g_mutex_init (&kernel->dma_mutex);

// In gst_dsp_kernel_finalize():
g_mutex_clear (&kernel->dma_mutex);

// Wrap DMA+send/recv sequences:
g_mutex_lock (&kernel->dma_mutex);
// ... DMA write, send, recv, DMA read ...
g_mutex_unlock (&kernel->dma_mutex);

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1786 @@
...
+  GST_INFO_OBJECT (kernel,
+      "  Duration: %u:%02u.%03u (%.6f seconds)",
+      audio_minutes, audio_seconds, audio_milliseconds, audio_duration_sec);
+
+
+  gst_buffer_unmap (buf, &map_info);
+
+  /* Create new memory for the audio output (only final_output_samples) */
+  GstMemory *new_mem = gst_memory_new_wrapped ((GstMemoryFlags) 0,
+      audio_out, final_output_samples * sizeof (gint16), 0, output_bytes,
```
v-singh1: check if audio_out is freed in all scenarios

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1786 @@
...
+      trim_start_samples = lo_sample;
+      trim_end_samples = kernel->chunk_samples - hi_sample;
+      GST_INFO_OBJECT (kernel,
+          "ISTFT: Chunk %zu (MIDDLE): Keep [%zu:%zu], trim both start=%zu end=%zu",
+          chunk_idx, lo_sample, hi_sample, trim_start_samples,
+          trim_end_samples);
+    }
+
+    if (trim_start_samples > 0 || trim_end_samples > 0) {
+      gsize trimmed_samples =
```
v-singh1: The trimming logic calculates trimmed_samples but doesn't validate bounds before calling memmove(), add check 
if (trimmed_samples <= 0) {
    GST_WARNING_OBJECT (kernel,
        "ISTFT: Trim would result in zero or negative output");
    g_free (audio_out);
    gst_buffer_unmap (buf, &map_info);
    return GST_FLOW_OK;  // or GST_FLOW_ERROR depending on policy
}

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+    if (tvm->final_output) {
+      delete (NDArray *) tvm->final_output;
+    }
+    tvm->final_output = new NDArray (output_array);
+
+    return GST_FLOW_OK;
+
+  }
+  catch (const std::exception & e)
+  {
```
v-singh1: If in-process TVM inference throws an exception, only a generic message is logged.  Doesn't clean up tvm->final_output, and doesn't indicate whether error was in set_input(), run(), or get_output(). Makes debugging hard. Wrap each step separately

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+
+  if (!needs_restart) {
+    g_free (wanted);
+    return;
+  }
+
+  g_print ("[TVM] Model changed, restarting tvm-model-daemon for: %s\n",
+      wanted);
+
+  if (system ("mkdir -p /var/lib/tvm_inference") != 0) {
```
cshilwant: NACK. Two issues:                                                                                                                                                                                                                  
                                                                                                                                                                                                                                   
1. **Filesystem modifications from a GStreamer plugin are not acceptable**. This is not the right layer for that - please revisit the approach entirely.                                                                               

2. Also, the existing implementation assumed root access. Production systems run services as non-root, password-protected users. With this logic, the plugin will block waiting for a password prompt - that is a hard runtime failure in any real deployment.                                                                                                                                                                                                                       
                                                                                                                                                                                                                       
Please rework both before resubmitting.

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+    GST_WARNING_OBJECT (tvm, "[TVM] Failed to create model cache directory");
+  }
+
+  if (!g_file_set_contents (TVM_MODEL_CACHE_FILE, wanted, -1, NULL)) {
+    GST_WARNING_OBJECT (tvm, "[TVM] Failed to write model cache file: %s",
+        TVM_MODEL_CACHE_FILE);
+  }
+
+  if (system ("systemctl restart tvm-model-daemon") != 0) {
```
cshilwant: Why is this needed?

`system("systemctl restart …")` blocks pipeline state change for upto 60 secs (gsttitvm.cpp, gst_ti_tvm_ensure_daemon_model)                                                                                                        
                                                                                                                                                                                                                                     
This runs synchronously inside `gst_ti_tvm_start` (i.e. during the `NULL→READY` or `READY→PAUSED` state change). A 60-second blocking poll is unacceptable - the pipeline bus will report an async timeout and the app may kill the pipeline. Use g_spawn_async + GLib's event loop polling, or at minimum move the restart to a separate thread. Replace system() calls with g_spawn_command_line_sync() which also gives proper error output.

Please drop this logic from plugin.

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+    }
+
+    GST_INFO_OBJECT (tvm,
+        "[TVM] Model daemon unavailable, loading model in-process");
+
+    gchar *lib_path = g_strdup_printf ("%s/deploy_lib.so", tvm->model_path);
+    gchar *param_path =
+        g_strdup_printf ("%s/deploy_param.params", tvm->model_path);
+
+    Module lib = Module::LoadFromFile (lib_path);
```
cshilwant: Looks like there will be memory leak?
Assume `LoadFromFile` throws an exception, the outer catch block fires but `lib_path` and `param_path` are never freed right?

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+{
+  struct TvmDaemonHeader req_hdr, resp_hdr;
+  gsize payload_bytes = input_size * sizeof (gfloat);
+
+  req_hdr.magic = TVM_DAEMON_MAGIC;
+  req_hdr.type = TVM_DAEMON_MSG_INFER_REQ;
+  req_hdr.len = (uint32_t) payload_bytes;
+
+  GST_INFO_OBJECT (tvm,
+      "[TVM] Daemon request: %zu floats (%zu bytes), input[0..3]=%.6f,%.6f,%.6f,%.6f",
```
cshilwant, line 876: These fire on every audio inference window at INFO level. At 16kHz/256-sample hop this is ~62 log lines/second. 

Better to downgrade to `GST_DEBUG_OBJECT`?

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+  }
+
+  if (!gst_ti_tvm_daemon_read_full (tvm->daemon_fd, tvm->daemon_output_buf,
+          resp_hdr.len)) {
+    GST_ERROR_OBJECT (tvm, "[TVM] Failed to read inference response payload");
+    return GST_FLOW_ERROR;
+  }
+
+  GST_INFO_OBJECT (tvm,
+      "[TVM] Daemon response: %zu floats (%u bytes), output[0..3]=%.6f,%.6f,%.6f,%.6f",
```
cshilwant, line 945: These fire on every audio inference window at INFO level. At 16kHz/256-sample hop this is ~62 log lines/second.

Better to downgrade to `GST_DEBUG_OBJECT`?

meson_options.txt
```diff
@@ -5,6 +5,9 @@ option('doc', type : 'feature', value : 'disabled', yield : true, description :
 option('profiling', type : 'feature', value : 'disabled', yield : true, description: 'Enable profiling building')
 option('dl-plugins', type : 'feature', value : 'enabled', yield : true, description: 'Enable DL Plugins')
 option('enable-tidl', type : 'feature', value : 'enabled', yield : true, description: 'Enable TIDL Offload')
+option('tvm-plugins', type : 'feature', value : 'disabled', yield : true, description: 'Enable TVM inference plugin (titvm)')
```
cshilwant, line 8: Do we really need two butvm-plugins (feature type) and enable-tvm (boolean type) are both required to compile gsttitvm.cpp. 

There's no documented scenario where `tvm-plugins=enabled` + `enable-tvm=false` is valid (correct me if there's any). Collapse to a single tvm-plugins feature

<IIRC, I had left the same comment in Draft PR>

meson.build
```diff
@@ -63,6 +76,45 @@ plugin_deps = [
...
+# tvm-plugins is only supported on am62d 
+if get_option('tvm-plugins').enabled() and SOC != 'am62d'
+  error('MESON_FAIL tvm-plugins is only supported on am62d (got SOC=', SOC, ')')
+endif
+
+if get_option('tvm-plugins').enabled() and get_option('enable-tvm')
+  tvm_runtime_lib = c.find_library('tvm_runtime', required : true)
+
+  # Pass TVM include paths as compiler flags
+  tvm_includes = [
```
cshilwant: wouldn't `tvm_runtime` library (which is already added as dependency with `tvm_runtime_lib`) already have these headers?

If they don't then let's not hardcode it...maybe try using find_program/pkg-config lookup

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1786 @@
...
+
+  /* Install user-facing properties */
+  g_object_class_install_property (gobject_class, PROP_MSG_TYPE,
+      g_param_spec_uint ("msg-type", "DSP Message Type",
+          "IPC message type (0x1020=STFT, 0x1030=ISTFT, 0x1040=De/Interleave)",
+          0, G_MAXUINT, DEFAULT_MSG_TYPE,
+          (GParamFlags) (G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS)));
+
+  g_object_class_install_property (gobject_class, PROP_PARAM2,
+      g_param_spec_uint ("param2", "Parameter 2",
```
cshilwant: param2 is an undescriptive property name.
The property encodes deinterleave/interleave direction for DSP_OP_DEINT_INTERLEAVE.

Please rename it for clarity in gst-inspect output

utility/gsttirpmsgctx.h
```diff
@@ -0,0 +1,101 @@
+/*
```
cshilwant, line 1: I see some formatting inconsistency in this file.
The inline functions use 4-space indentation while the rest of the GStreamer codebase (and this file's struct definition) uses 2-space.

In general for all 4 commits in this PR, run through gst-indent => https://pypi.org/project/gst-indent/

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1289 @@
...
+    GST_WARNING_OBJECT (tvm, "[TVM] Failed to write model cache file: %s",
+        TVM_MODEL_CACHE_FILE);
+  }
+
+  if (system ("systemctl restart tvm-model-daemon") != 0) {
+    GST_WARNING_OBJECT (tvm, "[TVM] Failed to restart tvm-model-daemon");
+  }
+
+  for (i = 0; i < 60; i++) {
+    g_usleep (G_USEC_PER_SEC);
```
cshilwant: Why is this needed? Looks like a HACK needed to wait until `tvm-model-daemon` is ready....gstreamer plugin shouldn't have such HACKs

cshilwant (changes requested):
FYI, NNStreamer already ships a TVM plugin with CPU and GPU hardware acceleration support - it's worth using that as a reference for GStreamer plugin conventions & best practices as you continue reworking on your custom plugin. 

You can also checkout [1] [2]  

[0] - https://github.com/nnstreamer/nnstreamer/blob/main/ext/nnstreamer/tensor_filter/tensor_filter_tvm.cc

[1] - https://gstreamer.freedesktop.org/documentation/plugin-development/?gi-language=c

[2] - https://docs.qualcomm.com/doc/80-70017-15B/topic/general-guidelines-for-developing-gstreamer-plugin.html

cshilwant:
please create a PR on `develop` branch.

meson.build
PrathamTI: c.find_library('tvm_runtime', ...) only locates the .so binary; it has no mechanism to discover header paths. I'll update the logic by trying pkg-config first and only falling back to the hardcoded paths if no tvm.pc is found.

meson_options.txt
PrathamTI: OK I'll keep a single `tvm-plugins` feature

ext/ti/gsttitvm.cpp
PrathamTI: Ok will remove

ext/ti/gsttitvm.cpp
PrathamTI: Will remove this logic.

ext/ti/gsttitvm.cpp
PrathamTI: Noted. Will remove this logic.

ext/ti/gsttitvm.cpp
PrathamTI: Thanks for catching this. Will update this.

ext/ti/gstdspkernel.cpp
PrathamTI: ok I'll rename it.

ext/ti/gsttitvm.cpp
v-singh1: Please check with patch i provided, current logic is not correct

PrathamTI: Addressed, maintaining a separate function to handle model switch: https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-9224a93535fc39b379c3d0bb066d021213855d11ed9446a8d10398964aa70035R816

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+  /* Allocate DMA buffers */
+  int r1 = dmabuf_heap_init ((char *) "linux,cma", kernel->input_buf_size,
+      kernel->rproc_device, &kernel->dma_input);
+  int r2 = dmabuf_heap_init ((char *) "linux,cma", kernel->output_buf_size,
+      kernel->rproc_device, &kernel->dma_output);
+
+  if (r1 != 0 || r2 != 0) {
+    GST_ERROR_OBJECT (kernel, "DMA alloc failed (r1=%d r2=%d)", r1, r2);
+    if (r1 == 0)
+      dmabuf_heap_destroy (&kernel->dma_input);
```
cshilwant, line 701: this only frees the input right? I see DMA output buffer still leaks when input alloc fails.

When `r1 != 0` but `r2 == 0` (output alloc succeeded), the output DMA buffer is silently leaked.

its `fd` and mapped memory are never freed. `dma_allocated` is set only at line 707 (after this block), so stop() won't catch it either.

```diff
  if (r1 == 0) dmabuf_heap_destroy (&kernel->dma_input);                                                                                                                                                                             
  if (r2 == 0) dmabuf_heap_destroy (&kernel->dma_output);
```

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+      trim_end_samples = kernel->chunk_samples - hi_sample;
+      GST_INFO_OBJECT (kernel,
+          "ISTFT: Chunk %zu (MIDDLE): Keep [%zu:%zu], trim both start=%zu end=%zu",
+          chunk_idx, lo_sample, hi_sample, trim_start_samples,
+          trim_end_samples);
+    }
+
+    if (trim_start_samples > 0 || trim_end_samples > 0) {
+      gsize trimmed_samples =
+          out_written_samples - trim_start_samples - trim_end_samples;
```
cshilwant, line 1194: I see a potential heap overflow here. gsize is unsigned. If trim_start_samples + trim_end_samples > out_written_samples (possible if the DSP returns fewer frames than expected), the subtraction wraps to a huge value, and `memmove` writes far past the audio_out allocation. 

Maybe add a guard before the subtraction.

```sh
// Must add before line 1193:                                                                                                                                                                                                      
  if (trim_start_samples + trim_end_samples > out_written_samples) {                                                                                                                                                                 
      GST_ERROR_OBJECT (kernel, "ISTFT: trim exceeds output, dropping chunk");                                                                                                                                                       
                                                                                                                                                                                                                                     
      // handle error                                                                                                                                                                                                                
                                                                                                                                                                                                                                     
  }    
```

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+  PROP_SELECTED_MODEL,
+  PROP_MODEL_ELEMS,
+  PROP_MODEL_PATH,
+  PROP_OVERLAP_FRAMES,
+  PROP_CHUNKING_MODE,
+};
+
+/* RPMsg/DMA infrastructure defaults */
+#define DEFAULT_RPROC_DEVICE    "/dev/remoteproc0"
+#define DEFAULT_RPROC_ID        8
```
cshilwant, line 138: Instead of hardcoding `rproc_id` and `remote_ep` to fixed values, can we retrieve these from an rpmsg header file? That way, any change to the rproc ID or remote core endpoint in firmware won't require a manual update to the GStreamer plugin.

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+
+/* STFT transform: accumulate audio. Overlap-save models process at EOS
+ * via dsp_kernel_process_chunks(); plain-windowing models process each
+ * complete window immediately via dsp_kernel_process_stream_window(). */
+static GstFlowReturn
+dsp_kernel_transform_stft (GstDspKernel * kernel, GstBaseTransform * trans,
+    GstBuffer * buf)
+{
+  GstMapInfo map_info;
+  if (!gst_buffer_map (buf, &map_info, GST_MAP_READWRITE)) {
```
cshilwant, line 974: Do we really need `GST_MAP_READWRITE` here? It forces a buffer copy if the buffer is shared or read-only (e.g., coming from a `filesrc`). I think `GST_MAP_READ` should suffice - could you double check?

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+  }
+  GST_OBJECT_UNLOCK (kernel);
+}
+
+static void
+gst_dsp_kernel_finalize (GObject * object)
+{
+  GstDspKernel *kernel = GST_DSP_KERNEL (object);
+  g_free (kernel->rproc_device);
+  g_free (kernel->model_path);
```
cshilwant, line 578: Any reason `gst_dsp_kernel_finalize` doesn't free all heap allocations?

The function only frees `rproc_device` and `model_path`. If the element is finalized after `stop()` wasn't called (rare but possible in error paths), `input_buffer` and `collected_audio` leak.

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
...
+    kernel->collected_audio_size += final_output_samples;
+    kernel->chunks_received++;
+
+    GST_INFO_OBJECT (kernel,
+        "ISTFT: ===== CHUNK %zu/%zu COLLECTED =====", chunk_idx + 1, n_chunks);
+    GST_INFO_OBJECT (kernel,
+        "ISTFT: Chunk %zu/%zu collected: %zu samples trimmed (from %zu original)",
+        chunk_idx + 1, n_chunks, final_output_samples, out_written_samples);
+
+    /* Calculate running total duration and expected chunk timing */
```
cshilwant, line 1234: why is sample rate **hardcoded** to 16000 Hz in duration logging?  In `gstdspkernel.cpp`, Line no. 1236, 1244, 1292, 1344

ext/ti/gsttitvm.h
```diff
@@ -0,0 +1,168 @@
...
+ * EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef __GST_TI_TVM_H__
+#define __GST_TI_TVM_H__
+
+#include <gst/gst.h>
+#include <gst/base/gstbasetransform.h>
+
+#ifdef __cplusplus
```
cshilwant, line 70: isn't the `extern "C" + G_BEGIN_DECLS nesting` redundant?

`G_BEGIN_DECLS` already expands to `extern "C" {` in C++ builds. 
The surrounding `#ifdef __cplusplus extern "C" {` wrapping is redundant.

cshilwant (changes requested):

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,1804 @@
+/*
```
cshilwant, line 1: `gstdspkernel.cpp` - zero buffer copy violations                                                                              
                                                                                                                       
Every DSP operation has two copies: GStreamer buffer → DMA input, then DMA output → GStreamer buffer.

```md
  ┌───────────────┬────────────────────────┬──────┐        
  │   Operation   │     Copy direction     │ Line │                                                                                                                                                                                           
  ├───────────────┼────────────────────────┼──────┤                                                                                                                                                                                           
  │ STFT          │ GstBuffer → DMA input  │ 912  │                                                                                                                                                                                           
  ├───────────────┼────────────────────────┼──────┤                                                                                                                                                                                           
  │ STFT          │ DMA output → GstBuffer │ 943  │                                                                                                                                                                                           
  ├───────────────┼────────────────────────┼──────┤                                                                                                                                                                                           
  │ ISTFT         │ GstBuffer → DMA input  │ 1086 │        
  ├───────────────┼────────────────────────┼──────┤                                                                                                                                                                                           
  │ ISTFT         │ DMA output → GstBuffer │ 1120 │                                                                                                                                                                                           
  ├───────────────┼────────────────────────┼──────┤                                                                            
  │ Deinterleave  │ GstBuffer → DMA input  │ 1415 │
  ├───────────────┼────────────────────────┼──────┤                                                                    
  │ Deinterleave  │ DMA output → GstBuffer │ 1450 │                                                                    
  ├───────────────┼────────────────────────┼──────┤                                                                    
  │ STFT chunking │ GstBuffer → DMA input  │ 1570 │                                                                    
  ├───────────────┼────────────────────────┼──────┤                                                                             
  │ STFT chunking │ DMA output → GstBuffer │ 1611 │                                                                                                                                                                                           
  └───────────────┴────────────────────────┴──────┘                                                                    
                                                                                                                       
```

the GStreamer buffers on both sides are always system-heap allocations - the DMA buffer is a separate allocation that will never be the same memory.

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1294 @@
+/*
```
cshilwant, line 1: In this plugin as well, I see `2–3` CPU copies per inference                                                                                                                                                                                                        

  1. Daemon path: write() GStreamer buffer → socket, read() → daemon_output_buf, then memcpy → output GstBuffer. 3 copies total. The socket IPC is structurally zero-copy-hostile.                                                             
  2. In-process path: CopyFromBytes GStreamer buffer → TVM NDArray, CopyToBytes NDArray → output GstBuffer. 2 copies.   
  3. No propose_allocation/decide_allocation hooks. gst_buffer_new_allocate(NULL,...) on every output buffer. 

```md
  Daemon path (3 copies):                                                                                                                                                                                                                     
                                                                                                                       
  ┌────────┬──────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                                                    
  │  Copy  │ Line │                                                     What happens                                                     │                                                                                                    
  ├────────┼──────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Copy 1 │ 882  │ gst_ti_tvm_daemon_write_full(fd, input_data, payload_bytes) — input floats written into socket                       │                                                                                                    
  ├────────┼──────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Copy 2 │ 938  │ gst_ti_tvm_daemon_read_full(fd, tvm->daemon_output_buf, ...) — result read from socket into intermediate heap buffer │                                                                                                    
  ├────────┼──────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤                                                                                                    
  │ Copy 3 │ 562  │ memcpy(out_map.data, tvm->daemon_output_buf, out_bytes) — intermediate buffer → GstBuffer                            │                                                                                                    
  └────────┴──────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘                                                                                                    
                                                                                                                                                                                                                                              
  In-process path (2 copies):                                                                                                                                                                                                                 
                                                                                                                       
  ┌────────┬──────┬─────────────────────────────────────────────────────────────────────────────┐                                                                                                                                             
  │  Copy  │ Line │                                What happens                                 │                                                                                                                                             
  ├────────┼──────┼─────────────────────────────────────────────────────────────────────────────┤                      
  │ Copy 1 │ 1008 │ input_array.CopyFromBytes(input_data, ...) — GstBuffer → TVM NDArray on CPU │                                                                                                                                             
  ├────────┼──────┼─────────────────────────────────────────────────────────────────────────────┤                      
  │ Copy 2 │ 565  │ out->CopyToBytes(out_map.data, out_bytes) — TVM NDArray → GstBuffer         │                      
  └────────┴──────┴─────────────────────────────────────────────────────────────────────────────┘ 
```
gst_buffer_new_allocate(NULL, output_size, NULL) allocates output from the system heap allocator. No DMA pool is proposed or accepted, so there is no shared physical memory region that could eliminate any of these copies.

cshilwant (changes requested):

ext/ti/gstdspkernel.cpp
PrathamTI: updated. Added `gst_dsp_kernel_propose_allocation` and ` gst_dsp_kernel_decide_allocation` Can you recheck?
STFT : GstBuffer → DMA input and STFT chunking: GstBuffer → DMA input is still present because wavparse hands buffers that were never allocated from dma_input's pool in first place. So the STFT input side theres nothing that can make that copy without proper upstream fix .

ext/ti/gsttitvm.cpp
PrathamTI: Added `gst_ti_tvm_propose_allocation` and `gst_ti_tvm_decide_allocation`

PrathamTI: Added `gst_ti_tvm_propose_allocation` and `gst_ti_tvm_decide_allocation`

PrathamTI:
Attached logs for Audio Classification (using YAMNET and VGGISH) and Audio Enhancement (using GCRN) pipelines:

1. Yamnet Audio Classification: https://gist.github.com/PrathamTI/25be66f2f6a0ca4e98fb742d1aa42e13
2. VGGISH Audio Classification: https://gist.github.com/PrathamTI/7d68060e3e2be747a00efcb890f12560
3. GCRN Audio Enhancement: https://gist.github.com/PrathamTI/fa369f6b639161408a5044b956067e53

ext/ti/gstdspkernel.cpp
PrathamTI: finalize() explicilty calls stop() [0] itself, regardless if the stop() function was called before. This adds finalize() as second, independent caller so even if the stop() process is skipped, finalize() call we call it instead, executing the same if() loops to free `input_buffer` and `collected_audio`.

If stop is run normally, this second call is a safe no-op - every field is gaurded.



[0] https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-edaedbac48f5e50d6978408eb60ecff745ae844cb774a05ece96f68c743bad62R642

ext/ti/gstdspkernel.cpp
PrathamTI: fixed by adding set_cap() function to retrieve `frame_rate`

ext/ti/gsttitvm.h
PrathamTI: Thanks for catching this, removed now

ext/ti/gstdspkernel.cpp
PrathamTI: Changed to `GST_MAP_READ`. 
Thanks

ext/ti/gstdspkernel.cpp
PrathamTI: Directly fetching rproc_id from rpmsg header [0]

[0] https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-edaedbac48f5e50d6978408eb60ecff745ae844cb774a05ece96f68c743bad62R201

ext/ti/gstdspkernel.cpp
PrathamTI: This code path doesn't exist anymore — dsp_kernel_transform_istft_chunked() was rewritten to write each batch's result directly into collected_dma at its final offset, so there's no audio_out heap buffer, no trim_start_samples/trim_end_samples, and no memmove anywhere in the current ISTFT path.

The underlying risk you're flagging is still handled, every batch checks resp.output_frame != frames_in_batch and fails loudly (GST_FLOW_ERROR, no attempt to proceed with a mismatched count). There's also a second check at stream completion comparing collected_dma_written against the precomputed collected_final_samples before handing anything downstream. So a short DSP response now aborts the chunk/stream instead of feeding a bad value.

meson.build
```diff
@@ -63,6 +76,52 @@ plugin_deps = [
...
+
+  # Add json-c dependency for shape auto-detection from deploy_graph.json
+  json_c_dep = dependency('json-c', required : true, method : 'pkg-config')
+  plugin_deps += [json_c_dep]
+  message('json-c library found - titvm will support shape auto-detection')
+endif
+
+# stft-plugins on am62d: needs ti_rpmsg_dma library for RPMsg/DMA functions
+if get_option('stft-plugins').enabled()
+  ti_rpmsg_dma_lib = c.find_library('ti_rpmsg_dma', required : true)
```
cshilwant, line 121: Why is this only included when `stft-plugins` is enabled?

I see `gsttitvm.cpp` includes `rpmsg.h` and `dmabuf.h` [0]
These symbols (`init_rpmsg`, `cleanup_rpmsg`, `dmabuf_heap_init`, `dmabuf_sync`, etc.) all come from `libti_rpmsg_dma` right? 

If yes, that library is added to `plugin_deps` only inside the `stft-plugins` block.
Enabling tvm-plugins without stft-plugins produces a clean compile followed by a linker failure with unresolved rpmsg/dmabuf symbols. The fix is to gate `ti_rpmsg_dma` on tvm-plugins as well.

[0] - https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-9224a93535fc39b379c3d0bb066d021213855d11ed9446a8d10398964aa70035R84

meson.build
```diff
@@ -63,6 +76,52 @@ plugin_deps = [
  math_dep
  ]
 
+# tvm-plugins is only supported on am62d
+if get_option('tvm-plugins').enabled() and SOC != 'am62d'
```
cshilwant, line 80: Similar guard is missing for `stft-plugins`. Add the same guard:                           
```sh
if get_option('stft-plugins').enabled() and SOC != 'am62d'                                                                                    
  error('MESON_FAIL stft-plugins is only supported on am62d (got SOC=', SOC, ')')                                                                                                                                                  
endif
```

ext/ti/meson.build
```diff
@@ -1,41 +1,80 @@
...
+
+if get_option('stft-plugins').enabled()
+  gst_ti_plugin_headers += [
+    '../../utility/gsttirpmsgctx.h',
+    'gstdspkernel.h',
+  ]
+endif
+
+# Include utility directory for gsttirpmsgctx headers
+utility_inc = include_directories('../../utility')
```
cshilwant: Why are we adding `include_directories` for all platforms?
This is needed only for `gsttirpmsgctx` which is applicable only when `stft-plugins=enabled`

But at line 77 in `ext/ti/meson.build`, `utility_inc` is unconditionally added to the compile flags of every translation unit in the plugin, even when `stft-plugins=disabled` and `gsttirpmsgctx.h` is never used. Gate it:                                                     
```sh
  if get_option('stft-plugins').enabled()                                                                                                                                                                                            
    utility_inc = include_directories('../../utility')                                                                                                                                                                                                                                                                                  
  endif
```

meson.build
```diff
@@ -63,6 +76,52 @@ plugin_deps = [
...
+    tvm_dep = declare_dependency(
+      compile_args : tvm_includes,
+      dependencies : [tvm_runtime_lib]
+    )
+    message('No tvm.pc found - falling back to hardcoded TVM include paths')
+  endif
+  plugin_deps += [tvm_dep]
+
+  # Add json-c dependency for shape auto-detection from deploy_graph.json
+  json_c_dep = dependency('json-c', required : true, method : 'pkg-config')
```
cshilwant, line 114: json-c is only used by `gsttitvm.cpp`, but adding it to `plugin_deps` affects all SoCs.

cshilwant:
> Attached logs for Audio Classification (using YAMNET and VGGISH) and Audio Enhancement (using GCRN) pipelines:
> 
>     1. Yamnet Audio Classification: https://gist.github.com/PrathamTI/25be66f2f6a0ca4e98fb742d1aa42e13
> 
>     2. VGGISH Audio Classification: https://gist.github.com/PrathamTI/7d68060e3e2be747a00efcb890f12560
> 
>     3. GCRN Audio Enhancement: https://gist.github.com/PrathamTI/fa369f6b639161408a5044b956067e53


Two small observations from the test logs that are worth noting:

1. `interleave_direction=0` uses underscore in your test logs [0] , not hyphen -- the property is registered as "interleave-direction" but the pipeline writes interleave_direction=0. GStreamer silently maps underscores to hyphens so it works, but it's inconsistent with every other property in these same pipelines.

2. YAMNet/VGGish pipelines omit `fft-size` and rely entirely on `model-path` sub-string detection (containing "yamnet" or "vggish") to auto-set `model_elems=64`. That means the model-path sub-string matching from the earlier review is actually load-bearing for those pipelines, not just a convenience. If the model artifacts directory is renamed or restructured without yamnet/vggish in the path (which may happen at customer's filesystem), the gstremaer element silently falls back to the GCRN formula `(fft_size/2+1)*2)` for YAMNet/VGGish. Is this behavior expected?
    
[0] - https://gist.github.com/PrathamTI/fa369f6b639161408a5044b956067e53

cshilwant (changes requested):

meson.build
PrathamTI: I verified again gsttitvm.cpp doesn't need ti_rpmsg_dma, it had #include "rpmsg.h" / #include "dmabuf.h" left over, but nothing in the file calls anything from either header. Removed the dead includes. That resolves the linker-failure case you flagged, at the actual source.

meson.build
PrathamTI: Added guard here as well: https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R80

ext/ti/meson.build
PrathamTI: Updated: https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-3d81c3014da9a750bb2d58a3fc14d3b4147d275afd02a21b4bb70f7209727ed3R72

meson.build
PrathamTI: it's already inside if get_option('tvm-plugins').enabled() [0] , and that block is already gated to am62d only by the SOC guard right above it [1]. So by the time json-c resolves, SOC==am62d is already guaranteed; I don't think it affects other SoCs. Am I missing anything?

[0] https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R89

[1] https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R80

PrathamTI:
> > Attached logs for Audio Classification (using YAMNET and VGGISH) and Audio Enhancement (using GCRN) pipelines:
> > ```
> > 1. Yamnet Audio Classification: https://gist.github.com/PrathamTI/25be66f2f6a0ca4e98fb742d1aa42e13
> > 
> > 2. VGGISH Audio Classification: https://gist.github.com/PrathamTI/7d68060e3e2be747a00efcb890f12560
> > 
> > 3. GCRN Audio Enhancement: https://gist.github.com/PrathamTI/fa369f6b639161408a5044b956067e53
> > ```
> 
> Two small observations from the test logs that are worth noting:
> 
> 1. `interleave_direction=0` uses underscore in your test logs [0] , not hyphen -- the property is registered as "interleave-direction" but the pipeline writes interleave_direction=0. GStreamer silently maps underscores to hyphens so it works, but it's inconsistent with every other property in these same pipelines.
> 2. YAMNet/VGGish pipelines omit `fft-size` and rely entirely on `model-path` sub-string detection (containing "yamnet" or "vggish") to auto-set `model_elems=64`. That means the model-path sub-string matching from the earlier review is actually load-bearing for those pipelines, not just a convenience. If the model artifacts directory is renamed or restructured without yamnet/vggish in the path (which may happen at customer's filesystem), the gstremaer element silently falls back to the GCRN formula `(fft_size/2+1)*2)` for YAMNet/VGGish. Is this behavior expected?
> 
> [0] - https://gist.github.com/PrathamTI/fa369f6b639161408a5044b956067e53

Thanks for catching these points:
1. Verified GObject treats -/_ as interchangeable in property lookups regardless (g_object_class_find_property() resolves both to the same GParamSpec), which is why it worked silently. Will use the hyphenated form in examples going forward. [0] 
2. Fixed this, now if model-path is set but doesn't match a known model and model-elems wasn't also set explicitly, gst_dsp_kernel_calculate_buffer_sizes() now fails loudly instead of falling through. [1]

[0] https://docs.gtk.org/gobject/class.ParamSpec.html
[1] https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/changes#diff-edaedbac48f5e50d6978408eb60ecff745ae844cb774a05ece96f68c743bad62R450

meson.build
cshilwant: ahh, I didn't see the `if get_option('tvm-plugins').enabled()`.
LGTM!

cshilwant (approved):
LGTM now!

Puyush45 (approved):

meson.build
```diff
@@ -63,6 +76,57 @@ plugin_deps = [
...
+  error('MESON_FAIL tvm-plugins is only supported on am62d (got SOC=', SOC, ')')
+endif
+
+# stft-plugins is only supported on am62d
+if get_option('stft-plugins').enabled() and SOC != 'am62d'
+  error('MESON_FAIL stft-plugins is only supported on am62d (got SOC=', SOC, ')')
+endif
+
+if get_option('tvm-plugins').enabled()
+  tvm_pkgconfig_dep = dependency('tvm', required : false, method : 'pkg-config')
```
sinha-shreyash: any specific reason we are preferring the pkg-config way explicitly for figuring out the dependencies? Why it cannot be the same as other dependencies listed above where it automatically searches for it behind the scenes ?

meson.build
sinha-shreyash: why not move this up in the block starting from line no 56

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1539 @@
...
+    return TRUE;
+
+  }
+  catch (const std::exception & e)
+  {
+    GST_ERROR_OBJECT (tvm, "TVM initialization failed: %s", e.what ());
+    return FALSE;
+  }
+}
+
```
v-singh1, line 986: The gst_ti_tvm_daemon_read_full & gst_ti_tvm_daemon_write_full , repeatedly call blocking read() and write() without a deadline. If the daemon remains connected but stops responding, the plugin can wait indefinitely during handshake, model loading, or inference. Closing the connection only in stop() is insufficient when shutdown is waiting for the blocked streaming operation.
Use nonblocking transfers with a configurable whole-transaction deadline and cancellation. Partial transfers must not restart the deadline. Discard a connection after an interrupted exchange so leftover payload bytes cannot corrupt the next request.

ext/ti/gsttitvm.cpp
```diff
@@ -0,0 +1,1539 @@
...
+  return TRUE;
+}
+
+static gboolean
+gst_ti_tvm_daemon_write_full (gint fd, gconstpointer buf, gsize size)
+{
+  gsize total = 0;
+
+  while (total < size) {
```
v-singh1, line 1045: Writing after the peer closes its connection can raise SIGPIPE. With the default signal disposition, the entire application terminates before the plugin can return an error.

ext/ti/gstdspkernel.cpp
```diff
@@ -0,0 +1,2164 @@
...
+  /* No-op: fd ownership stays with dma_buf_params. */
+}
+
+static void
+gst_dsp_kernel_allocator_class_init (GstDspKernelAllocatorClass * klass)
+{
+  GstAllocatorClass *allocator_class = GST_ALLOCATOR_CLASS (klass);
+
+  allocator_class->alloc = GST_DEBUG_FUNCPTR (gst_dsp_kernel_allocator_alloc);
+  allocator_class->free = GST_DEBUG_FUNCPTR (gst_dsp_kernel_allocator_free);
```
v-singh1: The allocator overrides the parent cleanup with an empty function
gst_dsp_kernel_allocator_free (GstAllocator * allocator, GstMemory * memory)
{
  /* No-op: fd ownership stays with dma_buf_params. */
}
GST_FD_MEMORY_FLAG_DONT_CLOSE already prevents the parent allocator from closing the externally owned fd. The parent still needs to release the GstFdMemory wrapper, clear its mutex, and clean up any remaining mapping. Skipping that cleanup leaks the wrapper whenever its last reference is released. Repeated pool-buffer discard and replacement can make the leak grow during streaming.

utility/gsttirpmsgctx.cpp
```diff
@@ -0,0 +1,157 @@
...
+      GST_INFO ("Reusing RPMsg channel (rproc=%u ep=%u fd=%d) refcount=%d",
+          rproc_id, remote_ep, c->fd, g_atomic_int_get (&c->refcount));
+      chan = c;
+      goto done;
+    }
+  }
+
+  /* No existing channel — open a new one. */
+  {
+    int fd = init_rpmsg ((int) rproc_id, (int) remote_ep);
```
sinha-shreyash: Are you using any custom wrapper on top of rpmsg ? any resistance towards using existing ti-rpmsg-char library?

meson.build
PrathamTI: Fixed,
These changes were during initial local implementation, updated meson build to use default dependency lookup.

ext/ti/gstdspkernel.cpp
PrathamTI: Removed gst_dsp_kernel_allocator_free

ext/ti/gsttitvm.cpp
PrathamTI: Added a new block: gst_ti_tvm_daemon_wait

ext/ti/gsttitvm.cpp
PrathamTI: Fixed: write() is now send(....,MSG_NONSIGNAL) so a closed peerr returns PIPE and buffers fails instead of SIGPIPE.

v-singh1 (approved):
Please check the comments,

cshilwant:
@PrathamTI please update commit [008ceb7](https://github.com/TexasInstruments/edgeai-gst-plugins/pull/388/commits/008ceb7d182aa3a6988a1fd7a1d7a0257ab12cd8) to add verified signature

